PR #1, opened 2022-11-21
ProfileImage view and user profile

Bonney:
Beginning to skeleton out some UI


Dimillian:
Would love if this code could be connected to the actual app. I would like to not merge code that is only preview / sample but which is actually used in current state of the app. Even if this is not connected to current data but just displaying fake data.

Also, for some reason this code does not show up in Xcode, you might want to move it inside the project.

PR #2, opened 2022-12-26
Update Nuke

kean:
- Update to 11.5.1 
- A couple of minor usage adjustments


kean:
Btw, `LazyImage` has basic video playback support. Not sure it recognizes `gifv` though. I'll look into it.

Dimillian:
Thanks! Ah, I should try to use it instead of my custom VideoPlayerView then. BTW, if you go on user profile with gif as user profile, Nuke actually take lot of time to load it in, you can try in the app here: https://mastodon.social/@dlx

Thanks for the PR and update!

PR #5, opened 2022-12-30
Color Set added for light and dark mode. close #3

### LINKED ISSUE #3: Provide a default light theme
Right now you can make your own in the settings, but I'll work on making a default one.

PrajeetShrestha:
Added ColorSet. 
- Added Dark Set 
- Added Light Set

### DIFF
--- a/Packages/DesignSystem/Sources/DesignSystem/ColorSet.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/ColorSet.swift
@@ -1,0 +1,26 @@
+import SwiftUI
+
+public protocol ColorSet {
+    var tintColor: Color { get set }
+    var primaryBackgroundColor: Color { get set }
+    var secondaryBackgroundColor: Color { get set }
+    var labelColor: Color { get set }
+}
+
+public struct DarkSet: ColorSet {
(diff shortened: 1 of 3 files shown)


Dimillian:
Damn I saw that just after I've actually pushed min for #3 :/ It looks slick! But I've decided to go with native colors in assets catalog.

PrajeetShrestha:
If you want to added number of themes, not just light and dark, I think this approach would work.

Dimillian:
Reverted my commit, I think the colors are not saved properly in AppStorage, basically setting light mode, and then relaunching the app = glitch. Will investigate later on your PR.

PrajeetShrestha:
> Reverted my commit, I think the colors are not saved properly in AppStorage, basically setting light mode, and then relaunching the app = glitch. Will investigate later on your PR.

I also experienced lot of glitches in the simulator, tried in a real device and it worked smoothly. Seems like AppStorage don't work with simulator as expected. 
Here's the reference: https://stackoverflow.com/questions/72400863/swiftui-appstorage-not-always-persisting-value

Dimillian:
Let's roll with that, I've tried multiple time to reproduce what I saw last night and I couldn't. 
Going forward I think we can remove the "prefer dark scheme" toggle and back it into themes. 
So we can provides different light and dark themes.

PR #7, opened 2022-12-30
AccountDetailView: support display name with emoji in navbar title

angristan:
I kept the same offset before displaying the name.

I removed the `title` property from `AccountDetailViewModel` because it's not used anymore. Since the `displayNameWithEmojis` method returns a view, I'm not sure if we should still go through the ViewModel somehow or if using `currentAccount.account?.displayNameWithEmojis` is OK?

| Before | After |
| ------ | ----- |
| SCR-20221230-v7n  | SCR-20221230-v6f|


Dimillian (approved):
Super nice, I totally forgot you can set any view as title using this!

PR #8, opened 2022-12-30
AppAccountView: support emoji in account name + slight design change

angristan:
Since `account.displayNameWithEmojis` took more height on the line, it looked funky, so I simplified the 3 lines into 2, by showing the full account address.

| Before | After |
| ------ | ----- |
| SCR-20221230-v9h  |  |


angristan:
(btw, no hard feeling if you discard these PRs, I'm just having fun because it's been a while since I touched SwiftUI 😇)

Dimillian (approved):
I love it, LGTM! Keep coming with them :)

PR #9, opened 2022-12-30
Mute video player by default

angristan:
It's pretty surprising when there is a meme video down the timeline playing at full volume  🥲


Dimillian (approved):
Yes I guess it's safer for now

PR #10, opened 2022-12-31
Remove macOS as destination from Xcode project

DD-P:
Package dependency TextView is iOS-only, so project will not successfully build on macOS. This small PR removes Mac as a supported destination in the IceCubesApp target.


Dimillian:
Unfortunately, macOS is planed, and I'll work out the TextView compatibility :)

PR #14, opened 2022-12-31
Add avatar position setting

jdanthinne:
This allow choosing the avatar position (top, as default, or leading)


Dimillian (approved):
This is a nice settings, the full width was a design choice, but make it closer to other apps is nice too. This is nicely done too. 
I'll eventually move the theme part into a sub screen as more settings are added, with a nice status preview on top.

PR #20, opened 2023-01-02
Add option to re-hide spoiler content after expanding

btylerh7:
Uses the same logic as the "Show more" button, but in reverse. This allows a user to hide the content that they expanded.

I am not sure if there is a better way to implement the conditional logic, but that was the simplest method I could come up with!

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusRowView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowView.swift
@@ -114,6 +114,16 @@ public struct StatusRowView: View {
         Text(status.spoilerText)
           .font(.body)
       }
+      if viewModel.displaySpoiler == false && !viewModel.status.spoilerText.isEmpty {
+        Button {
+          withAnimation {
+            viewModel.displaySpoiler = true
+          }
+        } label: {
+          Text("Show less")
+        }
+          .buttonStyle(.bordered)
+      }
       if viewModel.displaySpoiler {
         Button {
           withAnimation {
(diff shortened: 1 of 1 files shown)


Packages/Status/Sources/Status/Row/StatusRowView.swift
```diff
@@ -114,6 +114,16 @@ public struct StatusRowView: View {
         Text(status.spoilerText)
           .font(.body)
       }
+      if viewModel.displaySpoiler == false && !viewModel.status.spoilerText.isEmpty {
+        Button {
+          withAnimation {
+            viewModel.displaySpoiler = true
+          }
+        } label: {
+          Text("Show less")
```
Dimillian, line 123: I think you can use the same button for the show less / more, with something like 
```
  Button {
          withAnimation {
            viewModel.displaySpoiler.toggle()
          }
        } label: {
          Text(viewModel.displaySpoiler ? "Show less" : "Show more")
        }
```

And then only the condition above `!viewModel.status.spoilerText.isEmpty` can remains.

btylerh7: That looks a lot cleaner! So something like this?

```swift
if !viewModel.status.spoilerText.isEmpty {
  Text(status.spoilerText)
    .font(.body)
  Button {
    withAnimation {
      viewModel.displaySpoiler.toggle()
    }
  } label: {
      Text(viewModel.displaySpoiler ? "Show more" : "Show less")
  }
    .buttonStyle(.bordered)
}
if !viewModel.displaySpoiler { // Remove the "if/else" statement and make it just an if
  Text(status.content.asSafeAttributedString)
    .font(.body)
    .environment(\.openURL, OpenURLAction { url in
        routeurPath.handleStatus(status: status, url: url)
    })
  // Rest of code
}
```

Dimillian: Yup. It should just work :)

btylerh7:
Should be all good now!

PR #21, opened 2023-01-02
Allow tapping avatar to go details page

EvilOne:
When the avatar position is set to `leading`, we lose the ability to go to a user's details when tapping the avatar. This PR brings this functionality back.


Dimillian:
Thanks!

PR #22, opened 2023-01-02
Fix list visibility in light theme

EvilOne:
| Before | After |
| ------ | ----- |
| IceCubes-Before-AccountView-Lists | IceCubes-After-AccountView-Lists |

Theme being used:
CleanShot 2023-01-02 at 17 41 05@2x


Dimillian:
Good catch!

PR #23, opened 2023-01-02
Less typing padding constants

EvilOne:
Having to type `DS.Constants` when auto-complete fails to work in Xcode is too much! 😉 So, let's help our hands out by having to type less.

In this PR, I propose to move the constants from the `enum` and put them as a part of a `CGFloat`.

**Before:**  `.padding(.horizontal, DS.Constants.layoutPadding)`
**After:** `.padding(.horizontal, .layoutPadding)`


Dimillian:
Fair enough, previous pattern come from the UIKit world haha.

PR #24, opened 2023-01-03
Removed picker and replaced with themepreview view for selecting theme

PrajeetShrestha:
Added custom theme selector view.

### DIFF
--- a/Packages/DesignSystem/Sources/DesignSystem/Views/ThemePreviewView.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/Views/ThemePreviewView.swift
@@ -9,41 +9,24 @@ public struct ThemePreviewView: View {
   public init() {}
   
   public var body: some View {
-    VStack {
-      HStack {
-        Text("Select Theme")
-          .font(.title3)
-          .fontWeight(.medium)
-          .foregroundColor(theme.tintColor)
-        Spacer()
-        Button {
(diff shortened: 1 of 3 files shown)


Packages/DesignSystem/Sources/DesignSystem/Views/ThemePreviewView.swift
```diff
@@ -0,0 +1,109 @@
...
+  private let gutterSpace: Double = 8
+  @EnvironmentObject private var theme: Theme
+  @Environment(\.dismiss) var dismiss
+  
+  public init() {}
+  
+  public var body: some View {
+    VStack {
+      HStack {
+        Text("Select Theme")
```
Dimillian, line 14: I would be in favour of using a NavigationStack with a title and a toolbar item as a close button. Only because it make it consistent with all other screens in the app. Unless this is hard push for your own style there (which is fine too) :)

PrajeetShrestha: I agree. Will update.

Dimillian (approved):
Loving it! Thank you so much for the great theme support we have now!

PrajeetShrestha:
> Loving it! Thank you so much for the great theme support we have now!

Thanks for initiating this wonderful project :)

PR #25, opened 2023-01-03
Made filtered key optional

PrajeetShrestha:
When filtered key is missing from the response data. 

App can't load the timeline data. So made it optional.

### DIFF
--- a/Packages/Models/Sources/Models/Status.swift
+++ b/Packages/Models/Sources/Models/Status.swift
@@ -38,7 +38,7 @@ public protocol AnyStatus {
   var visibility: Visibility { get }
   var poll: Poll? { get }
   var spoilerText: String { get }
-  var filtered: [Filtered] { get }
+  var filtered: [Filtered]? { get }
 }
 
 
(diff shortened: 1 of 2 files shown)


Dimillian:
I never know which data is optional or not... Thanks! Wonder why it was working on a fresh account with no filters tho.

PrajeetShrestha:
I have not studied the Mastodon APIs in detail as well, but I got something like this for filtered key as well. 

timeline parse error: keyNotFound(CodingKeys(stringValue: "voted", intValue: nil), Swift.DecodingError.Context(codingPath: [_JSONKey(stringValue: "Index 7", intValue: 7), CodingKeys(stringValue: "poll", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"voted\", intValue: nil) (\"voted\").", underlyingError: nil))

This is for **voted** key now. I keep getting these logs when user is not in signed in state. Also see lot of this in timeline. 

Simulator Screen Shot - iPhone 14 Pro - 2023-01-03 at 20 05 29

EvilOne:
Thanks for fixing! I too was getting that message from the last update and was going to look into it!

Dimillian:
Saw that too, pushed a fix.

PR #28, opened 2023-01-04
Fix missing app icon 2

EvilOne:
Again, this works, but please test in TestFlight.

This time I:
- Deleted the "bad" icons
- Re-added the "bad" icons
- I reverted the build setting of "Include all App Icon assets" to NO
- Manually added all the custom icons within "Alternate App Icons"

CleanShot 2023-01-04 at 06 54 57@2x

image


EvilOne:
One should be able to use the Include All App Icon Assets setting in the build, but I am at a loss right now because it works on my machine! 😉

Dimillian:
Let's try again.

PR #29, opened 2023-01-04
fix: restore tap behavior on avatar when position is leading

jdanthinne:


jdanthinne:
Sorry, didn't see it was already fixed in your main branch…

PR #30, opened 2023-01-04
Avatar shape

jdanthinne:
This adds a setting to choose the avatar shape (rounded, as currently, or circle).


Dimillian (approved):
Nice settings, I love it!

PR #33, opened 2023-01-06
Fix crash on message tab-iPad only

EvilOne:
When running only on the iPad, switching to the Messages tab caused a crash with this error:

CleanShot 2023-01-06 at 09 44 17@2x

This PR fixes this. Tested on iPhone and iPad.


Dimillian:
Is it a random crash? It's done the same way in all other tab, this is weird, hmmm

EvilOne:
I could only get it to crash on the iPad in landscape mode for this tab.

Dimillian:
Yeah I confirm it crash and this fix works but this is mystic lol

PR #34, opened 2023-01-06
Fix loading local time line

EvilOne:
It appears that the API can return
- a valid URL
- NULL
- and **BLANK**

The last case above was causing the json decoder to throw when trying to create a valid URL.

| Before | After |
| ------ | ----- |
| CleanShot 2023-01-06 at 12 15 39@2x | CleanShot 2023-01-06 at 12 14 19@2x |

### DIFF
--- a/Packages/Models/Sources/Models/Status.swift
+++ b/Packages/Models/Sources/Models/Status.swift
@@ -8,6 +8,15 @@ public struct Application: Codable, Identifiable {
   public let website: URL?
 }
 
+extension Application {
+  public init(from decoder: Decoder) throws {
+    let values = try decoder.container(keyedBy: CodingKeys.self)
+
+    name = try values.decodeIfPresent(String.self, forKey: .name) ?? ""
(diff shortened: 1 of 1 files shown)


Packages/Models/Sources/Models/Status.swift
```diff
@@ -8,6 +8,15 @@ public struct Application: Codable, Identifiable {
   public let website: URL?
 }
 
+extension Application {
+  public init(from decoder: Decoder) throws {
+    let values = try decoder.container(keyedBy: CodingKeys.self)
+
+    name = try values.decodeIfPresent(String.self, forKey: .name) ?? ""
+    website = try? values.decodeIfPresent(URL.self, forKey: .website)
```
EvilOne, line 16: If decode cannot create a valid URL, just assign it nil.

EvilOne: Example response for status

CleanShot 2023-01-06 at 12 20 18@2x

Dimillian:
Nice catch!

PR #35, opened 2023-01-06
Skip unsupported notifications

angristan:
As an instance admin one can receive `admin.sign_up` and `admin.report` notifications ([mastodon code](https://github.com/tootsuite/mastodon/blob/55bef1e34fc3b07ed7f762d565a161e74e128016/app/javascript/mastodon/actions/notifications.js#L135-L136)).

image

Since IceCubesApp does  not support them, we see a bunch of dividers in the app. This MR skips them entirely.

| Before | After |
| ------ | ----- |
|  SCR-20230106-qtq |SCR-20230106-qt8|


Dimillian:
Thanks for the fix and I'll be sure to support them soon!

PR #36, opened 2023-01-07
Update StatusRowContextMenu to include a "Copy Text" option

btylerh7:
I am learning Japanese, and there are many times where I need to copy all or parts of a post to look up words or put them through a translator. I thought it would be a nice addition to add this type of feature to the app. Some Mastodon apps make the text of a post selectable in the status detail view (not sure how hard that would be), but I also have seen some that have a "Copy Text" option in a context menu. Because the app already has a context menu when you long press on a post, I figured that would be a great place to put it!

The way I have it, it appears to work well for most posts. The only issue is that sometimes I will get the links to hastags in with a post. for example, if I use this post:

Screenshot 2023-01-06 at 7 33 28 PM

I will get this a a response:

```
I’m not at  {
}
#AnimeLA18 {
	NSLink = https://urusai.social/tags/AnimeLA18
}
 today, but I have some pictures of cosplayers from last night, so I’ll aim to post them today. First, a Chisato (left) and Takina (right) from  {
}
#LycorisRecoil {
	NSLink = https://urusai.social/tags/LycorisRecoil
}
 in their combat uniforms which are designed to look like school uniforms. The face masks they’re wearing matches their characters’ hair: white for Chisato and black for Takina {
}
```
I think this is just a byproduct of using `asSafeAttributedString`, so I'm not sure if there is a way to get around that. However, I still think it does the job and will be fine for most use cases. If you have any suggestions on making it more efficient, I'd be happy to make changes!


Dimillian:
I think you could try `asRawText` on it, it should parse out the HTML and leave you with the text only.

btylerh7:
That works! I also don't need the template literal that way!

Dimillian:
All good! Thanks :)

PR #37, opened 2023-01-07
Added a share button into status context menu

jan-patrick:
Added the option to share a status via the context menu (added the button “Share this post”):
Simulator Screen Shot - iPhone 14 - 2023-01-07 at 14 52 04

It uses the built-in share function of iOS and allows sharing a post as many iOS users are used to – like me sending e.g. an interesting status to friends on iMessage.

Please let me know if I missed anything – this is my first time touching SwiftUI in a long while.


Dimillian:
I like it. Thanks!

PR #44, opened 2023-01-08
Support In-App Safari

divadretlaw:
Add option to open links with In-App-Safari instead of system Safari.

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
+++ b/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
@@ -31,7 +31,6 @@ struct SettingsTabs: View {
       .toolbarBackground(theme.primaryBackgroundColor, for: .navigationBar)
       .withAppRouteur()
       .withSheetDestinations(sheetDestinations: $routeurPath.presentedSheet)
-      .withSafariRouteur()
     }
     .onAppear {
       routeurPath.client = client
@@ -41,6 +40,7 @@ struct SettingsTabs: View {
         await currentInstance.fetchCurrentInstance()
       }
     }
+    .withSafariRouteur()
     .environmentObject(routeurPath)
   }
   
--- a/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
+++ b/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
@@ -31,6 +31,7 @@ struct SettingsTabs: View {
       .toolbarBackground(theme.primaryBackgroundColor, for: .navigationBar)
       .withAppRouteur()
       .withSheetDestinations(sheetDestinations: $routeurPath.presentedSheet)
+      .withSafariRouteur()
     }
     .onAppear {
       routeurPath.client = client
@@ -40,6 +41,7 @@ struct SettingsTabs: View {
         await currentInstance.fetchCurrentInstance()
       }
     }
+    .environmentObject(routeurPath)
   }
   
   private var accountsSection: some View {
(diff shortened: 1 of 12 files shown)


Dimillian:
It's breaking adding new account for now

EvilOne:
The `StatusCardView` doesn't appear to honour the in-app setting when tapped.

image


https://user-images.githubusercontent.com/169561/211198853-33e0f92a-91e6-48eb-9c4a-2af171b3c1bd.mp4

divadretlaw:
> It's breaking adding new account for now

@Dimillian Using `UIApplication.shared.open(url)` to avoid being captured by In-App Safari and restore previous behavior.

> The StatusCardView doesn't appear to honour the in-app setting when tapped.

@EvilOne Fixed, the SafariRouteur needed to be attached to the `NavigationStack` not the content of it. Otherwise the environment is not propagated to the navigation link destinations.

IceCubesApp/App/Tabs/Settings/SettingsTab.swift
```diff
@@ -82,6 +84,9 @@ struct SettingsTabs: View {
       NavigationLink(destination: remoteLocalTimelinesView) {
         Label("Remote Local Timelines", systemImage: "dot.radiowaves.right")
       }
+      Toggle(isOn: $preferences.useInAppSafari) {
```
Dimillian, line 87: Could you make it a Picker instead? It'll make the label shorter and also more clear like "Browser: In App Safari, System Safari" as choices.

Dimillian: On a small iPhone simulator, the label does not fit on one line.

EvilOne:
> > The StatusCardView doesn't appear to honour the in-app setting when tapped.
> 
> Fixed, the SafariRouteur needed to be attached to the `NavigationStack` not the content of it. Otherwise the environment is not propagated to the navigation link destinations.

Moving the `SafariRouteur` to the `NavigationStack` solves the issue and makes total sense! Do we need the two extensions to find the topmost VCs, though? I have the code prior to your changes and just moved the `.withSafariRouteur()` to the `NavigationStack` level, and all worked. I must be missing something.

divadretlaw:
> Moving the `SafariRouteur` to the `NavigationStack` solves the issue and makes total sense! Do we need the two extensions to find the topmost VCs, though? I had the code before your changes and just moved the `.withSafariRouteur()` to the `NavigationStack` level, and all worked. I must be missing something.
> 
> Thanks, and sorry for all the questions.

@EvilOne thank you for testing the PR, questions are always welcome 😄 

Found another issue and resolved it by finding the top-most view controller, since you can only present another view controller from the currently presenting view controller.

The issue was on Account Details, when opening a link within the "About Fields". Since they are opened as a sheet they prevented the `SFSafariViewController` to be presented, because the view controller was no longer the top most one. Now the `SFSafariViewController` should be presented always from the top most view controller.

IceCubesApp/App/Tabs/Settings/SettingsTab.swift
divadretlaw: @Dimillian done, let me know if the implementation of it is ok.

EvilOne:
@divadretlaw Good catch and thanks for the explanation!

Dimillian:
Thanks for all this. I'll take time to do a good round of testing tomorrow before merging.

Dimillian:
Let's have it in now, it seems all good, had little time to test in the end! Thanks a lot!

PR #46, opened 2023-01-08
Minor improvements

divadretlaw:
The About Fields Sheet does not have a close button

FieldClose

Mentions should also support `-` since Mastodon instances may contain it

| Before | After |
|:-:|:-:|
| MentionPattern_before | MentionPattern_after | 

The checkmark of the selected account is transparent, but should probably be white on green

| Before | After |
|:-:|:-:|
| SelectedAccount_before | SelectedAccount_after |

Empty Notification Text is sometimes cut off

Notifications


Dimillian:
This is very solid thanks!

PR #59, opened 2023-01-11
Add button to re-hide sensitive media once expanded

btylerh7:
I just added a very simple button in the top left corner to re-hide the media. I tried the default button style and on some images like this one it was pretty hard to actually see the button, so I changed the button style to `.bordered`. It should only appear if the image is marked as sensitive and the user unhides it. Here is a screenshot:

Screenshot 2023-01-10 at 11 30 57 PM


btylerh7:
That looks a lot cleaner!

PR #61, opened 2023-01-11
Support Custom Emojis

divadretlaw:
Proof of concept rendering custom emojis using [EmojiText](https://github.com/divadretlaw/EmojiText)

Renders Markdown-Text and custom emojis with the format `:shortcode:` and supports local and remote emojis.
[Nuke](https://github.com/kean/Nuke) is used for loading and caching remote emojis

### DIFF
--- a/IceCubesApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/IceCubesApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/divadretlaw/EmojiText",
       "state" : {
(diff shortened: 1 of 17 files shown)


Dimillian:
WHOA, this is so cool!

divadretlaw:
Definitely room for improvement but I think it generally works

Dimillian:
What's left / not working you think? I've tried with and without this PR on the same feed for a while and it seems to render all the same + emojis.

divadretlaw:
Added some placeholders before loading and removed the custom emojis from the navigation title, since they will not render there at the moment.

The placeholder is `square.dashed` but can be customized via `.environment(\.placeholderEmoji, some UIImage)`

In my opinion it is mergable

Dimillian:
Nice will do a new round of test tomorrow before merging.

Dimillian:
Let's go! I can't thank you enough for this!

PR #69, opened 2023-01-12
Fix Link Handling

divadretlaw:
Resolving links to a status results in a lot of false positives, breaking opening said links with `SFSafariViewController`

Twitter links for example where wrongly identified as Mastodon Status Links since it only checked if the URL ends with an `id`

> Example of a link that generates a false positive: https://twitter.com/Dimillian/status/1608909053836808192

Therefore we should check if the URL is a known connection on the current Mastodon Instance and only then treat the URL as a Status URL.

Not really sure where to add this but currently it is added to `Client` and fetched when the `Client` is switched and user data is fetched.

Seems more like `Instance` information but currently the `Client` is available in the `Routeur` but maybe we move it to `Instance` and make it available within `Routeur`?

### DIFF
--- a/Packages/Env/Sources/Env/CurrentAccount.swift
+++ b/Packages/Env/Sources/Env/CurrentAccount.swift
@@ -22,12 +22,21 @@ public class CurrentAccount: ObservableObject {
   
   private func fetchUserData() async {
     await withTaskGroup(of: Void.self) { group in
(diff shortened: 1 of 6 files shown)


Dimillian:
Smart, thanks for this. I'll merge as in for now. It can definitely use improvement where we indeed sore the connection elsewhere.

PR #70, opened 2023-01-12
Polls

EvilOne:
Adds the ability to add a poll to a message

- Honours the number of poll entries and the maximum number of characters per option based on the instance the user is posting from
- The durations presented to the user are taken from my Mastodon instance
  - The API returns a minimum and a maximum for the duration

| Poll Creation | Poll Created |
| ------ | ----- |
|  |  |


Packages/Status/Sources/Status/Editor/PollView.swift
```diff
@@ -0,0 +1,128 @@
+import SwiftUI
+import DesignSystem
+import Env
+
+struct PollView: View {
```
scalbatty, line 5: Should this be named StatusEditorPollView for clarity, and be filed within the `Components` folder as the other accessories?

Packages/Status/Sources/Status/Editor/Components/StatusEditorAccessoryView.swift
```diff
@@ -14,6 +14,8 @@ struct StatusEditorAccessoryView: View {
   @FocusState<Bool>.Binding var isSpoilerTextFocused: Bool
   @ObservedObject var viewModel: StatusEditorViewModel
   @State private var isDrafsSheetDisplayed: Bool = false
+
+  @Binding var isShowingPoll: Bool
```
scalbatty, line 18: Why do we need a binding var here instead of just using the existing `viewModel.showPoll` flag?

Packages/Status/Sources/Status/Editor/Components/StatusEditorAccessoryView.swift
```diff
@@ -35,6 +38,15 @@ struct StatusEditorAccessoryView: View {
...
         }
+
+        Button {
+          withAnimation {
+            viewModel.showPoll.toggle()
+          }
+        } label: {
+          Image(systemName: "chart.bar")
+        }
+        .disabled(isShowingPoll)
```
scalbatty, line 49: Why is this disabled when showing the poll? Couldn't it also hide the poll?
It should be disabled if there are medias attached though.

scalbatty (commented):
I've had indexOutOfRange fatal errors when removing/adding options. Looking into the root cause…

Packages/Status/Sources/Status/Editor/Components/StatusEditorAccessoryView.swift
seanlifetime: The poll icon is disabled to help make the intention of creating a poll clear. However, we can run into an issue where the user taps the icon, fills in some information, taps it again to hide it then hits the post. 
- Was the intention to post with or without a poll? 
- Did they mean to dismiss or throw away the poll by hiding it? 
- Did they forget about the poll and mean to post it?

seanlifetime: Good catch for your second point. When I get back in front of my computer, I will update the code to disable this.

Packages/Status/Sources/Status/Editor/PollView.swift
seanlifetime: This sounds good to me. I will update this too!

Packages/Status/Sources/Status/Editor/Components/StatusEditorAccessoryView.swift
seanlifetime: Nice catch! 😿 Yes, you are right. Unfortunately, I moved my POC code into this code base, and it appears I forgot to change this to use the view model.

seanlifetime:
Open items to come back to:
- [ ] Disable the ability to add a poll when an attachment attached
- [ ] indexOutOfRange fatal errors when removing/adding options

EvilOne:
Open items to come back to:
- [ ] Disable the ability to add a poll when an attachment attached
- [ ] indexOutOfRange fatal errors when removing/adding options

EvilOne:
> I've had indexOutOfRange fatal errors when removing/adding options. Looking into the root cause…

@scalbatty I am having issues reproducing the indexOutOfRange. Would you happen to have some steps to reproduce this?

Dimillian:
I've removed the re focus on options delete for now, it was the cause of the index crash. I'll merge like this, because it's good enough and see if I can have the focus works differently. 

Thanks a lot for this!

PR #80, opened 2023-01-15
Implement Localization

### LINKED ISSUE #77: [Feature Request] Localization
Set up a system for localizing the app for languages other than English.

This will probably require at least the following points:
- [ ] Convert all hard-coded English text in the views etc. to localization keys
- [ ] Add the Localizable.strings file for English
- [ ] Possibly set up an easy-to-use collaboration tool (such as Crowdin)
- [ ] Adapt UI for right-to-left languages (I assume most of it is handled automatically by SwiftUI)

I volunteer to help setting all of this up and providing translations for German. It is just a very big change in many files and some work beyond that, so I didn't want to just open a PR.

vollkorntomate:
This PR lays the foundation for multi-language support (localization) in IceCube (#77).

I introduced a bunch of localization keys, that's why there are changes in almost all view-related files. I hope the key system makes a little sense, it should help new translators to easily find the correct translations.

Unfortunately, it is not too trivial to modularize the localization files. Each package would need a `.lproj` folder for each language (and you apparently need to create them manually, I didn't find any options in Xcode). However, this is definitively a thing I could still change and take the somewhat clumsy approach when adding new language.

Let me know what you think!

---

Closes #77

### DIFF
--- a/IceCubesApp.xcodeproj/project.pbxproj
+++ b/IceCubesApp.xcodeproj/project.pbxproj
@@ -64,6 +64,7 @@
 		9FD542E72962D2FF0045321A /* Lists in Frameworks */ = {isa = PBXBuildFile; productRef = 9FD542E62962D2FF0045321A /* Lists */; };
 		9FE151A6293C90F900E9683D /* IconSelectorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9FE151A5293C90F900E9683D /* IconSelectorView.swift */; };
 		9FE3DB57296FEFCA00628CB0 /* AppAccount in Frameworks */ = {isa = PBXBuildFile; productRef = 9FE3DB56296FEFCA00628CB0 /* AppAccount */; };
+		E9B576C329743F4C00BCE646 /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = E9B576C529743F4C00BCE646 /* Localizable.strings */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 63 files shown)


Dimillian:
I think this is a great start! But will take me time to review as I need to test everything that we don't display any keys instead of the actual string.

vollkorntomate:
> image

Thanks for noticing! I fixed this one and two others that I found at a closer look.

LorenAmelang:
So is this active in 1.1.2? Not a problem for me, but I'm really curious how it is supposed to work...  

I see "Translate" in a mysterious selection of posts: 
All English but maybe @pschiller or @m0bi13 triggered it for the entire conversation?
All English but from @muenchen.social?
All English but from a woman in Portugal, and including a reply from @0xabad1dea?
All English but from Francesco Bianco?
All English but from @IchHabRect?
All English but from @LauraAmalasunta?
And in every post from @leafs_s@mstdn.science !

And no "Translate" in a reply (to Dimillian) that is clearly all in French. 

Can you calm my curiosity?

vollkorntomate:
Hey @LorenAmelang,
this PR only added the ability to translate the app's interface, not the contents of posts.

As far as I know, the translate button is only shown for posts whose language is different from your default post language and translations done with this language as the target. Your default post language is not necessarily the language in which you view the app. For example, my phone and the app are set to German, so I see the app's interface in German. But I set my default post language to English, so now the translate button is only shown on non-English posts and posts get translated to English, not German.

You can check which language is set for you as your default post language by creating a new post in IceCubes and looking at which language is pre-selected there. Alternatively, check your settings in the Mastodon web client, you can also change it there (not yet possible in IceCubes).

As for replies, if the original post's language is set to English, the replies are also automatically pre-set to English. If someone then enters French text, they would need to change the post's language manually (there is no language detection based on the content).

To summarize, the main prerequisites in order for the "Translate" button to work properly are:
1. The user posting set the correct language for the content (e.g. selected French for a French text)
2. Your default post language is set to a language that you are comfortable with
3. The language of the post is different from the language you set
4. DeepL supports the source and target languages

Hope that clarified some things, let me know if you have any further questions!

LorenAmelang:
@vollkorntomate Thanks for the clues! 

I'm all set to English on my end. So if every @leafs_s@mstdn.science post (which are all in English) includes "Translate", he must have his default post language set to something else? Is there a way to see his default language setting from my end?

If one reply (all in English) to a post includes "Translate", does that mean the author of that reply has a different default post language? 

If every reply (all in English) to a post includes "Translate", what does that mean? 

So you are saying the "foreign-ness" of people's handles is irrelevant? (Except as a clue for me that they might have some other language default...) 

Could there be a way (in the Share Sheet?) to trigger translation when somebody posts in a language that isn't their default or my default?

vollkorntomate:
I just checked and it seems, all of @leafs_s@mstdn.science's posts are set to be Japanese (the small `JA` on the top-right.

image

In the Mastodon web client, the "Interface Language" corresponds to the default post language. So my guess is that this user has their interface set to Japanese and doesn't change a post's language to English when posting English content (many mobile apps, including the official Mastodon one, don't let you change a post's language).

You won't be able to find out a user's default post language, only the language that is set for a single post. This (currently) works best in the Web client, in my opinion. Maybe a small indicator will be added to IceCubes in the future.

LorenAmelang:
Aha! That "JA" in your screenshot does show up in the web version of my Neuromatch Social instance - they use the "Glitch" version of Mastodon. I've never seen it anywhere else. And I've never consciously noticed any non-English posts there, either, I only just checked because I recognized the double chevron or whatever that is called to the right of it. Love that bit of Glitch when they allow 10K chars per post! So now I have a new tool for exploring...

PR #83, opened 2023-01-16
Implement language selection for new posts

### LINKED ISSUE #76: [Feature request] Option to select post's language in the composer
It is important for multi-language people to indicate the language of their post.

The `/api/v1/statuses` endpoint accepts a `language` parameter. Its value should be a ISO 639 language code that represents the language of a status update. On the official Mastodon UI, this can be selected using the language icon.

IMG_0174

Ice Cubes could have a similar button in the New Post composer. It could be placed between the CW and Drafts buttons on the toolbar. The default language could be derived from the system language.

vollkorntomate:
This PR adds the ability to select a new post's language. It takes into account the preference set by the user (sent by the server). Replies take the language of the parent post by default.

Preview (the simulator is set to English as first and German as second language):
image

Closes #76

### DIFF
--- a/Packages/Models/Sources/Models/Status.swift
+++ b/Packages/Models/Sources/Models/Status.swift
@@ -50,6 +50,7 @@ public protocol AnyStatus {
   var spoilerText: String { get }
   var filtered: [Filtered]? { get }
   var sensitive: Bool { get }
+  var language: String? { get }
 }
 
 
@@ -83,6 +84,7 @@ public struct Status: AnyStatus, Codable, Identifiable {
   public let spoilerText: String
   public let filtered: [Filtered]?
   public let sensitive: Bool
+  public let language: String?
   
   public static func placeholder() -> Status {
     .init(id: UUID().uuidString,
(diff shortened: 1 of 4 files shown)


EvilOne:
I cannot get the globe icon to show when I post a message. Would this happen in only certain cases? I see the following:



And here is what the xCode console shows:
CleanShot 2023-01-16 at 18 05 54@2x

EvilOne:
But when I reply, I do get the language set and can set a different language.

Dimillian:
@EvilOne Fixing it, it's probably because your account don't have a default language, I'm setting the default to English. 
@vollkorntomate Good works! I'll do a few fixed and merge it! 

I think an improvement could be to show a sheet instead and add search into it. Would make it easier to browse languages.

Dimillian:
image

EvilOne:
@Dimillian  Good to know. Even with the latest TestFlight build I still don't see the language selection option for new posts. Is this expected behaviour?

Dimillian:
@EvilOne Are you sure? You should see the globe if no language is set hmmmm

EvilOne:
@Dimillian here is what I see. 

image

image

EvilOne:
Double checked to make sure I have the latest. image

EvilOne:
One last thing. I checked my settings for posts and it is set to automatically determine language of post.

Dimillian:
That's sooooo weird. My secondary account is like yours and it's showing the globe. Will investigate at some point if you don't find a fix first.

PR #85, opened 2023-01-17
Fix two issues related to share extension

myihsan:
This PR fixes two issues below.
- Missing `EnvironmentObject`
- Can't post when sharing images only


Dimillian:
Thanks for that!

PR #120, opened 2023-01-19
Allow to block/unblock and mute/unmute accounts

### LINKED ISSUE #118: Support blocking/unblocking and muting / unmuting account
It would be nice to support block/unblock of account directly in the app. Probably a new option in the context menu of accounts and toots.

https://docs.joinmastodon.org/methods/accounts/#block

rafaelcaricio:
Fixes #118


Dimillian:
Thanks a lot!

PR #124, opened 2023-01-19
Add localizations for German

vollkorntomate:
I translated the app to German and added some localization keys I somehow missed earlier.


Dimillian:
This is wonderful! My wife will do Ukrainian next and I'll do French!

PR #135, opened 2023-01-20
Make vapidKey in InstanceApp model optional

### LINKED ISSUE #16: Can't add account
I can't add the account for my instance. I enter the domain "mastodon.denibol.com" but it does nothing. It's a Pleroma backend but other Mastodon clients does not have any problem loggin in. Instance metadata on https://mastodon.denibol.com/api/v1/instance

japareaggae:
This allows Ice Cubes to log into GoToSocial instances.
Related to #16, but not a full fix since the issue is also about Pleroma support.


Dimillian:
Nice!

PR #203, opened 2023-01-20
Fix weird scrolling when opening status detail

vollkorntomate:
When tapping on a post that is beyond a certain height (including pictures, link previews etc.), the view would be scrolled a little bit down, hiding the profile picture and name. This PR fixes this – opening the detail view now looks smoother.

Here's a recording of the current behavior:

https://user-images.githubusercontent.com/38211057/213803328-53e09642-9872-4090-8f4a-b45e6e778ad5.mp4


Dimillian:
I'm not sure you can do that. Try open a post in the middle of a long context, it'll not scroll correctly to the focused one anymore.

vollkorntomate:
Oh, I see. I will try to do it differently then.

PR #205, opened 2023-01-20
Fix bug where App Icon in settings displays wrong image for some values.

jhalickman:
It looks like when converting from Icon name to Icon enum if the number is two digits only the last digit is pulled giving the wrong image. This changes how it converts the String `AppIconAlternate[XXX]` into an Icon enum.


Dimillian (approved):
Thanks much better

PR #210, opened 2023-01-21
Moved app version to the section footer

patricks:
Since the app version row doesn't have any action assigned I think it can also be moved to the section footer.

See Screenshot:
Screenshot 2023-01-21 at 08 24 13


Dimillian:
yes make sense! thanks. I like it.

PR #214, opened 2023-01-21
Fix bookmark/unbookmark text in context menu

vollkorntomate:
When you tapped on the three dots context menu of a bookmarked post, the label would still say "Bookmark" and not "Unbookmark".


Dimillian:
Nice catchè

Issue #99, opened 2023-01-18
Polls are not showing full sentence, finishing three dots


MariaFeodora:
![Image](https://user-images.githubusercontent.com/65338652/213212666-dd6fa61a-ffd7-4d1a-9da9-550307a4e2e5.png)

PR #226, opened 2023-01-21
Scale text for polls

EvilOne:
A possible solution for truncating text issue found in #99 

| Before | After |
| ------ | ----- |
|  |  |


Dimillian:
I think that should do it!, thanks.

PR #235, opened 2023-01-21
Fix typo in German localization

patrickhoefler:
`geboosted` → `geboostet`

These germanized English words always end in -t, never in -d. `Geboostet` isn't in the [Duden](https://www.duden.de/) yet, but it works exactly the same as e.g. [`gebootet`](https://de.wiktionary.org/wiki/Flexion:booten).


Dimillian:
Thanks!

patrickhoefler:
You're very welcome — and thank you for a great Mastodon client 🙂

Issue #225, opened 2023-01-21
Bug: typo in Dutch translation 


EmilJacobs:
The home page is translated to “beging” while it should be “begin” or perhaps “start” or maybe even “thuis”.

EvilOne:
@EmilJacobs I don't speak Dutch. But I set the language of another app to Dutch, and they have **Startpagina** would this work?

EmilJacobs:
Yes, that would work, if it is not too long?

EvilOne:
@EmilJacobs does this work for you?
<img height="50%" src="https://user-images.githubusercontent.com/169561/213918364-b9316407-b7b3-4dd1-a951-a3bf6f0d451b.png">

EmilJacobs:
I think so, yes 🙂

EvilOne:
Thanks, I have updated things, and it will be released in a future version.

icod:
@Dimillian, this issue may be closed. "beging" has been changed to "startpagina" and is now simply "start".

PR #239, opened 2023-01-21
Add localisation plurals

### LINKED ISSUE #198: Minor bug: "1 new posts" incorrectly pluralised
Hi! Love the app.

The new-post notification says "1 new posts" instead of "1 new post".

It seems to be here in the code: https://github.com/Dimillian/IceCubesApp/blob/997a33df697e940a5310b01e1304eace1a7be60f/Packages/Timeline/Sources/Timeline/TimelineViewModel.swift#L52

I found this about pluralisation in Xcode: https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals

There seem to be a few more always-pluralised strings in https://github.com/Dimillian/IceCubesApp/blob/997a33df697e940a5310b01e1304eace1a7be60f/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings.

EvilOne:
- Fixes #198, which adds a way to create plurals of words within the app.
- Also addresses #225


Dimillian:
Looking at the two other PR we'll have to handle Italian and Chinese in there too. I'll look at it tomorrow once everything else will be merged. Buts it's great!

EvilOne:
If you merge the other languages, I can quickly put them into this PR too

EvilOne:
No rush from you, I know it is late.

Dimillian:
All good, all other translations update are merged :)

EvilOne:
@Dimillian updated. It looks like #262 doesn't update the key I had removed ("timeline.n-new-posts %lld"), which is good and shouldn't cause conflict issues. With that said, I recommend merging this before #262 since it has more updated translations.

Dimillian:
Well let's go with this one!

EvilOne:
Sounds good to me! We have to start somewhere!

PR #240, opened 2023-01-21
Italian Translation

torre76:
Hi @Dimillian,

I hope you can include also Italian on available language.
Thank you for your outstanding work!

Regards
Gian Luca


Dimillian:
Thank!s

PR #241, opened 2023-01-21
Timeline VoiceOver Improvements

richkolasa:
### Timeline

- Added accessibility custom actions for available status options
- Accessibility actions also include the ability to open any @ mentions in the status
- Grouped status content so that VoiceOver users only need to swipe once between statuses (before/after video attached)
- I left the status detail page alone, so that images can be accessed (plus that screen is a bigger LOE)
- Here and there the code formatting doesn't look right...but looks fine in Xcode. Not sure what's going on there. Let me know if I need to fix that.

https://user-images.githubusercontent.com/19654210/213884202-5f8fe1c6-adda-45f0-ba26-ec99b4b33e5c.MP4

https://user-images.githubusercontent.com/19654210/213884200-19aaa385-3861-4d50-8fbf-c0fa71983d02.MP4

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusRowView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowView.swift
@@ -67,10 +67,31 @@ public struct StatusRowView: View {
           viewModel.displaySpoiler = false
         }
       }
-      .contextMenu {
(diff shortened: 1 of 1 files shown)


Dimillian:
This is awesome thanks a lot!

PR #251, opened 2023-01-21
Only ask for draft when text entered by user

### LINKED ISSUE #228: Feature Request: Delete draft reply w/o confirm if no text has been typed by me
I fairly often hit reply on a post and either decide that I want to check the other replies first or I just don’t want to reply anymore. But canceling is more steps than I think adds value. 

The current design is:
1. I tap “reply” on a post
2. I type nothing (although there is some auto populated text)
3. I tap “cancel”
4. I am presented the option to “delete draft” or “save draft”. I have to hit “delete draft”
5. I am taken back to the post from step 1.

I would rather the app skip step 4 and immediately delete the draft. I will lose nothing since I didn’t type anything. Even if I accidentally hit cancel, I will be right back at the post that I was replying to and can just tap that button again. Of course, if text was typed by the user in step 2, then the save prompt in step 4 is needed. 

The behavior I’m requesting is already present for original (non-reply) posts. I’m just asking for the same on reply posts. I suspect the intended behavior is to have the same for reply posts, but that the current code is looking for an empty text field and the auto-populated quote of the post being replied to is being counted as user typed text.

EvilOne:
Fixes #228.

### DIFF
--- a/Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
+++ b/Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
@@ -64,6 +64,16 @@ public class StatusEditorViewModel: ObservableObject {
     showPoll || !selectedMedias.isEmpty
   }
 
+  var shouldDisplayDismissWarning: Bool {
+    var modifiedStatusText = statusText.string.trimmingCharacters(in: .whitespaces)
+
+    if let mentionString, modifiedStatusText.hasPrefix(mentionString) {
+      modifiedStatusText = String(modifiedStatusText.dropFirst(mentionString.count))
+    }
+
+    return !modifiedStatusText.isEmpty && !mode.isInShareExtension
+  }
+
   @Published var visibility: Models.Visibility = .pub
 
   @Published var mentionsSuggestions: [Account] = []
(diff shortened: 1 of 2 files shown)


Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
```diff
@@ -72,9 +82,10 @@ public class StatusEditorViewModel: ObservableObject {
   private var currentSuggestionRange: NSRange?
 
   private var embeddedStatusURL: URL? {
-    return URL(string: embeddedStatus?.reblog?.url ?? embeddedStatus?.url ?? "")
+    URL(string: embeddedStatus?.reblog?.url ?? embeddedStatus?.url ?? "")
   }
 
+  private var mentionString: String?
```
Dimillian, line 88: This is never set or am I missing something?

Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
```diff
@@ -172,6 +183,10 @@ public class StatusEditorViewModel: ObservableObject {
       visibility = status.visibility
       statusText = .init(string: mentionString)
       selectedRange = .init(location: mentionString.utf16.count, length: 0)
+
+      if !mentionString.isEmpty {
```
Dimillian, line 187: Should this be reversed?

Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
seanlifetime: It gets set on line 188.

Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
Dimillian: Whoa I was not awake lol. Thanks

seanlifetime: This is checking that the local variable has data. When it does it sets the ivar mentionString.

Dimillian: Yes, I was confused by the name shadowing at a quick glance.

PR #256, opened 2023-01-22
Basic RTL statuses

orenro:
This PR displays statuses on the time line with the appropriate alignment for right-to-left languages like Arabic and Hebrew.

This is far from all the work that has to be done to make the app support mixed languages with different directionality, but it's a simple first step that does make timelines much easier to read in the most common case.

The code relies on the language field for the status being set correctly, which most servers seem to do. It does not handle mixed-language statuses, where different paragraphs require different directionality.


Dimillian:
This is a great start, thanks a lot of that!

mustafaalbazy:
Please add support for RTL typing when writing/creating a new post. 

Also, the current RTL support on the timeline doesn't seem to work every time. Is it based on the first character of the post? Or how does the function detect the language? 

Thanks.

orenro:
Directionality for writing posts was next on my list, but real life distracted me. Hopefully I'll have some time to implement it soon.

The code I submitted is a very naive implementation and relies on the language reported by the post itself: each Mastodon post has a metadata field saying what language it supposedly is in. This field might (and often is) wrong.

A better implementation would indeed be to put some language detection code in. Looking at the first character is what a lot of applications historically do, but that fails when the first word in an otherwise RTL post is in latin script. This is an annoying behavior that RTL users historically struggle to compensate for by deliberately forcing their text to start with an RTL word whether this makes for good phrasing or not.

iOS and macOS do have a built-in ML model for detecting languages, which is likely the best solution. I'll have to experiment with it to see if running it on every post is too much of a performance hit. If it is, a simple statistical character count threshold will also probably do a decent job.

Finally, it's probably not enough to do this for every post, but for every paragraph inside the post. I often see posts where different paragraphs should have different directionality.

PR #259, opened 2023-01-22
Some localization issues

WedgeSparda:
Hi, found some minor issues after updating to 1.0.5.

- Added missing translation on spanish. 
- Replacing hardcoded text with localization key (non spanish translations still missing).


Dimillian:
Thanks!

PR #262, opened 2023-01-22
Better Simplified Chinese Localization

nixzhu:
- Improved some translations.
- Added two more strings.

### DIFF
--- a/IceCubesApp/Resources/Localization/zh-Hans.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/zh-Hans.lproj/Localizable.strings
@@ -6,10 +6,10 @@
 "action.retry" = "重试";
 
 "alert.button.ok" = "OK";
-"alert.error" = "错误!";
+"alert.error" = "错误！";
 
-"placeholder.loading.long" = "加载, 加载, 再加载 ....";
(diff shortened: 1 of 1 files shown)


Jerry23011:
OK, so most fixes are with the punctuations, it won't make any big difference, but thanks for that anyway.

I would personally prefer `邮件` over `电邮` because we know `邮件` means email, and it is what Apple uses for their Mail app. If you really want to emphasize it's "Email", then perhaps try `电子邮件`? `电邮` is rather old-styled  :)

Secondly, `且能鼓励我完善它` doesn't sound like a native Chinese speaker talking to the users, would you please keep the original one, thanks~

IceCubesApp/Resources/Localization/zh-Hans.lproj/Localizable.strings
```diff
@@ -31,7 +31,7 @@
 
 // MARK: Instances
 "instance.info.domains" = "域名";
-"instance.info.email" = "电邮";
+"instance.info.email" = "邮件";
```
Jerry23011, line 34: Thank you so much :)

nixzhu:
@Jerry23011 新年快乐！The email's translation is updated.

Dimillian:
🚀 let's got!

PR #265, opened 2023-01-22
Setting to hide boost content

jdanthinne:
For users who prefers only original content (without boosts), this adds a setting to hide the boosts content in the timeline, with a button to reveal it.

Simulator Screen Recording - iPhone 14 Pro - 2023-01-22 at 13 53 27

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/DisplaySettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/DisplaySettingsView.swift
@@ -51,6 +51,7 @@ struct DisplaySettingsView: View {
           }
         }
         Toggle("settings.display.translate-button", isOn: $userPreferences.showTranslateButton)
+        Toggle("settings.display.show-boost-content", isOn: $userPreferences.showBoostContent)
       }
       .listRowBackground(theme.primaryBackgroundColor)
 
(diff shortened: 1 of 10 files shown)


henrik:
Interesting how it shows the toot but hides the content. I’ve been meaning to suggest a feature to filter out toots based on content. E.g. if I follow someone for their tech content but don’t want to see their sports content or whatever.

This kind of UI might be good for that too. I’m sure it’d avoid a lot of complexity that you’d run into if you tried to hide those toots altogether.

jdanthinne:
In fact I was hesitating between hiding the content (as provided) and completely removing the posts from the list (globally or per user).

patrickhoefler:
How would this relate to Mastodon's built-in ["hide boosts" feature](https://docs.joinmastodon.org/methods/accounts/#form-data-parameters-2)?

henrik:
Yeah, side effects from returning an empty view could be stuff like saying “2 new posts” but only showing 1, or showing none, etc.

jdanthinne:
> Yeah, side effects from returning an empty view could be stuff like saying “2 new posts” but only showing 1, or showing none, etc. 

If the choice is to completely hide boosts, I wouldn't do that with EmptyView but by filtering the data source instead.

Dimillian:
Make sense but `TimelineViewModel` is far from simple and probably time for a refactor if we need to inject user settings on top of that. But sound good to filter at the datasource level.

jdanthinne:
> How would this relate to Mastodon's built-in ["hide boosts" feature](https://docs.joinmastodon.org/methods/accounts/#form-data-parameters-2)?

You're right! This was s already built-in with Mastodon API, on a per-user basis. I'll try it use that.

jdanthinne:
The "API way" is here: https://github.com/Dimillian/IceCubesApp/pull/305

Issue #151, opened 2023-01-20
BUG: URL is counted as 23 characters


pacorob:
URL should be counted as 23 characters.

Dimillian:
There is a spec? Url don't count more than 23 characters ever? Nice!

pacorob:
> There is a spec? Url don't count more than 23 characters ever? Nice!

Yes there is: https://docs.joinmastodon.org/user/posting/#links

PR #281, opened 2023-01-22
Include spoiler text count when applicable

### LINKED ISSUE #276: Bug: characters in spoiler tag not included in post character count
I had a post written in Ice Cubes with a spoiler, and it wasn’t letting me post, so I copied the text and took it over to the basic web app. I pasted the text into the body and had close to 500 characters without the CW. When I posted the spoiler text into the CW field, it took me over 500 characters, though Ice Cubes had said I was under 500 even with the spoiler text.

EvilOne:
Fixes #276


Dimillian:
Wonderful thanks! There is this one #151 and I wonder if it can be done in almost the same way.

EvilOne:
I will look into that one too

PR #286, opened 2023-01-22
Adjusted file names to match content

kalkwarf:
Previously, a couple classes were renamed to correct spelling, but the file names were not. This PR fixes that.


Dimillian:
Nice!

PR #287, opened 2023-01-22
Primitive validation of domain before trying to connect

kalkwarf:
Every domain needs at least one "dot" and can't end with a "dot." Don't try connecting unless/until these are both true.


Dimillian:
THanks!

PR #292, opened 2023-01-22
Digest: Popular posts from timeline

### LINKED ISSUE #132: In-app algorithmic timeline
This could be another unique feature of IceCubes. Something like https://github.com/hodgesmr/mastodon_digest embedded in the app that could calculate and generate a timeline based on some configurable criteria.

rafaelcaricio:
Fixes  #132

I'm not sure that this is something you are interested in having in the Ice Cubes app, but maybe nice to consider. The "algorithm" is based on the https://github.com/hodgesmr/mastodon_digest project and uses a fixed configuration (equivalent to `SimpleWeighted` option and using the `home` timeline). If this is something you want to explore, then other options could be added in the user settings, or an entirely new set of views for the Digest feature could also be considered.

Mastodon API has a limit on how many toots are visible in the user's home timeline (400 toots). More details here: https://github.com/mastodon/mastodon/issues/2301 Based on that list, the code in this PR creates a digest of "popular posts".

I keep running the [mastodon_digest](https://github.com/hodgesmr/mastodon_digest) myself from time to time, and I thought it would be nice to have it embedded in a client like Ice Cubes.

As a side note, those are my first semi-complex lines of Swift. I'm happy to refactor and eager to learn how it could be better.

### DIFF
--- a/Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
+++ b/Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
@@ -67,6 +67,8 @@ class TimelineViewModel: ObservableObject, StatusesFetcher {
 
   func fetchStatuses(userIntent: Bool) async {
     guard let client else { return }
+    // TODO: Check if the timeline is TimelineFilter.digest
+    // TODO: fetch the statuses until specified time (last 24 hours) and then run the algorithm
     do {
       if statuses.isEmpty {
         pendingStatuses = []
--- a/Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
+++ b/Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
@@ -90,7 +90,7 @@ class TimelineViewModel: ObservableObject, StatusesFetcher {
         if timeline == TimelineFilter.digest {
           return
         }
-        var newStatuses: [Status] = await fetchNewPages(minId: first.id, maxPages: 20, maxStatuses: nil)
+        var newStatuses: [Status] = await fetchNewPages(minId: first.id, maxPages: 20)
         if userIntent || !pendingStatusesEnabled {
           pendingStatuses.insert(contentsOf: newStatuses, at: 0)
           statuses.insert(contentsOf: pendingStatuses, at: 0)
@@ -123,7 +123,7 @@ class TimelineViewModel: ObservableObject, StatusesFetcher {
       to: Date()
     )!
     let timestamp = ISO8601DateFormatter().string(from: earlyDate)
-    let allStatuses = await fetchNewPages(minId: timestamp, maxPages: nil, maxStatuses: 1000)
+    let allStatuses = await fetchNewPages(minId: timestamp, maxPages: 400)
 
     var originals: [Status] = []
     var boosts: [Status] = []
@@ -150,7 +150,7 @@ class TimelineViewModel: ObservableObject, StatusesFetcher {
     return originals.gatherPopular(percentile: 95) + boosts.gatherPopular(percentile: 95)
   }
 
-  func fetchNewPages(minId: String, maxPages: Int?, maxStatuses: Int?) async -> [Status] {
+  func fetchNewPages(minId: String, maxPages: Int) async -> [Status] {
     guard let client else { return [] }
     var pagesLoaded = 0
     var allStatuses: [Status] = []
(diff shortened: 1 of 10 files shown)


Dimillian:
This is amazing. I'll play around a bit but I love the idea! I do want to explore it and I do like the idea that the user can setup the weight himself. Let me look at the code but honestly this is very good I think we could merge it soon!

Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
```diff
@@ -100,6 +111,44 @@ class TimelineViewModel: ObservableObject, StatusesFetcher {
...
+    let hoursAgo = 24
+
+    // Get as many posts as possible (max seems to be 400 though https://github.com/mastodon/mastodon/issues/2301)
+    let earlyDate = Calendar.current.date(
+      byAdding: .hour,
+      value: -hoursAgo,
+      to: Date()
+    )!
+    let timestamp = ISO8601DateFormatter().string(from: earlyDate)
+    let allStatuses = await fetchNewPages(minId: timestamp, maxPages: 400)
```
Dimillian, line 126: So the API accept a minId that is in a form of a timestamp and not an id of a Status? 
Also 400 might be bit much considering that page contains 40 status? With 10 you would already get 400 status of data. Not sure there.

Dimillian:
image
I definitely get older than 24H status at the moment. Is that normal?

Dimillian:
General feedbacks: On my account it return around 10 status and it's hard to understand why they're sometimes. For example a random reply with 1 likes, I feel like I miss a lot of other popular status? I think a reason would be awesome, like a little header to explain (like Twitter) why it's there.

Dimillian:
https://user-images.githubusercontent.com/535509/213975196-1aab647f-aacb-48b6-bf27-a91f930237ef.mov

Snapshot from my account. Maybe it makes sense in the end tho!

Dimillian:
We can follow some user feedbacks here https://mastodon.cloud/@icecubesapp/109736950398919054

rafaelcaricio:
Hey, happy to hear this is of interest. I will try to clarify below some of my own observations.

There are many nuances with the Digest which might bring frustration in the experience. The frustration might be lowered with more clarity about the formula for calculating the digest, plus options to change the formula. An example using the same formula is this app: https://fediview.com/ (source code: https://github.com/adamghill/fediview)

- Accounts that I follow with low follower numbers gets a highlight even if they have 1 star.
- Account information might not be up-to-date, as that info is what my Mastodon instance have stored. Fediverse if very volatile in this aspect, servers are quickly out of date with remote accounts metadata like followers counts.
- The number of toots that appear in the Digest are (currently) posts above the 95th percentile of scores of all loaded posts.

rafaelcaricio:
I just noticed that the F.A.Q in this site clarifies some things: https://fediview.com/faq

Maybe the Ice Cubes should have something similar?

Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
rafaelcaricio: I just put "a huge number" there. But the Mastodon is limited to max 400 toots for the home timeline API call.

rafaelcaricio:
>  I definitely get older than 24H status at the moment. Is that normal? 

Yes, because if somebody retoots that or if that toot belongs to the 400 most recent from the home timeline.

Dimillian:
Not sure about the next step on this. Do you want to build the settings and explanation screen? I can also look at it a bit later in the week.

rafaelcaricio:
@Dimillian I will try to come up with some settings (like https://fediview.com/ ) and explanation for the Digest. I think it would be better to have this in the app before releasing, since people will ask a lot of the same questions.

Dimillian:
Yep exactly. Would sit fine in the setting you can do a new Form for the Digest.

Dimillian:
And I think at least a slider to choose the percentile would be nice.

rafaelcaricio:
@Dimillian I won't have time in the next few days to work on this. So, I just want to let you know if someone else wants to take over. I don't want to sit on this when other people might want to move it to the finish line.

joebarbere:
I was thinking about this earlier this week. Having so many posts to scroll through is becoming a chore 😝 A digest sounds like a great addition! I can try chipping away at the UI later this week. I'm new to SwiftUI so you may wind up getting to it before me, if so, no worries.

PR #298, opened 2023-01-22
Ignore twitter handle taps

EvilOne:
Temporarily ignore the tapping of Twitter handles so the app will stop crashing.

Twitter handle example: 
CleanShot 2023-01-22 at 17 22 48@2x


EvilOne:
I'm going to close this. I have a playground going with a possible fix that would allow tapping Twitter "links".

PR #305, opened 2023-01-23
Allow hiding or showing user boosts

jdanthinne:
This adds a button to show or hide user boost on a per-user basis, like notifications.
I've added a new button, but not sure if it is the correct place (Follow button is wrapped sometimes).

Simulator Screen Shot - iPhone 14 Pro - 2023-01-23 at 12 02 11

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailHeaderView.swift
+++ b/Packages/Account/Sources/Account/AccountDetailHeaderView.swift
@@ -157,8 +157,7 @@ struct AccountDetailHeaderView: View {
       }
       .foregroundColor(.gray)
       .font(.footnote)
-      .padding(.top, 4)
-      .padding(.bottom, 5)
+      .padding(.top, 6)
     }
   }
 }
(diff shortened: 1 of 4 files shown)


Dimillian:
Can we have it in the ... menu instead? Apart from that so cool!

Dimillian:
Or we need a find a better UX for the notifications and boost button. It make the follow button too thin right now.

jdanthinne:
I could move those two buttons below, but it takes space… Wherever we put them, I think they need to be together I think.

Simulator Screen Shot - iPhone 14 Pro - 2023-01-23 at 16 54 57

Dimillian:
Yeah I think it's fine as you'll have this layout only when you follow the person.

jdanthinne:
Moved the joined date below the name to fill up some empty space…

Simulator Screen Shot - iPhone 14 Pro - 2023-01-23 at 17 09 37

Dimillian:
That's very smart and it look close to perfect! Ready to merge?

PR #316, opened 2023-01-23
Fix twitter links

### LINKED ISSUE #102: Support twitter link
I see a lot of people doing username@twitter.com, when you taps these links the app will crash, maybe we can add some handling that if the link ends in Twitter.com it opens it up in safari

Image

EvilOne:
Fixes #102 

The example that I used in the video to test: https://arsenalfc.social/@BrewCityGunners/109734277436192767

https://user-images.githubusercontent.com/169561/214132360-fda08ec9-8c98-46cf-aa5f-f93e9052d725.mp4


Dimillian:
Let's merge that and see how it run for a while on my TF. 
I'm not a fan of altering the content of a status or adding compute time to decoding them. I still think we could handle them as you tap on it, but this solution is clean. 🚀

EvilOne:
Neither am I, seemed appropriate and straightforward, but your update is better!

Issue #317, opened 2023-01-23
Bug: Internal linking to hashtags is too aggressive


xurble:
It looks for the token "tags" anywhere in the path and then links to a hashtag that is the last token on the path.

So, for example, this toot:
https://mastodon.scot/@mattgemmell/109737860334286795 
has a link to this page:
https://mattgemmell.scot/books/once-upon-a-time/stories/tags/

This is incorrectly routed as a link to the hashtag page for "tags"

xurble:
Fix is released

PR #318, opened 2023-01-23
This is a fix for #317

xurble:
This is a fix for #317
Well more of an improvement than a fix.
Instead of looking for tags anywhere, we now look for /tags/TagName exactly.  This can still fail if it happens to be a link to somewhere not on the post creator's home server that follows that pattern but by the time the parsing is done the post creator isn't available without a lot of re-plumbing.
But anyway, this should be better.

### DIFF
--- a/Packages/Env/Sources/Env/Router.swift
+++ b/Packages/Env/Sources/Env/Router.swift
@@ -62,9 +62,9 @@ public class RouterPath: ObservableObject {
   }
 
   public func handleStatus(status: AnyStatus, url: URL) -> OpenURLAction.Result {
-    if url.pathComponents.contains(where: { $0 == "tags" }),
-       let tag = url.pathComponents.last
+    if url.pathComponents.count == 3 && url.pathComponents[1] == "tags"
     {
+      let tag = url.pathComponents.last!
       navigate(to: .hashTag(tag: tag, account: nil))
       return .handled
     } else if let mention = status.mentions.first(where: { $0.url == url }) {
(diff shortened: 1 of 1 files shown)


Packages/Env/Sources/Env/Router.swift
```diff
@@ -62,9 +62,9 @@ public class RouterPath: ObservableObject {
   }
 
   public func handleStatus(status: AnyStatus, url: URL) -> OpenURLAction.Result {
-    if url.pathComponents.contains(where: { $0 == "tags" }),
-       let tag = url.pathComponents.last
+    if url.pathComponents.count == 3 && url.pathComponents[1] == "tags"
     {
+      let tag = url.pathComponents.last!
```
kalkwarf, line 67: I'm not a maintainer here, but I find force unwraps to be _bad_. The previous code used `guard` to ensure the tag could be extracted.

Looking at the Issue this is set to fix, I'm uncertain why `count == 3` and `pathComponents[1]` are important preconditions. 

Can you please explain the logic to me?

xurble: The existing code matches for tags anywhere in the path so `/foo/tags/bar/` and `/tags/foo/bar/` and just `/tags/` are all matches.  Then it plucks the last part off the path and makes that the navigation target for the tag view.

However, only paths in the form `/tags/name_of_tag` are actually valid, so that's the test here.

3 pathComponents, namely `[ "/", "tags", "name_of_tag"]`. 
`pathComponent[1]` has to be `"tags"` or it's not a match, so not `/foo/tags` for example 
The force unwrap is guaranteed to work because we know there are 3 components so `last` will definitely have a value.

xurble: Having said that, I have no strong feelings about force unwraps, so I've changed it back to the style it was before.

Dimillian:
Thanks for that and the review!

PR #327, opened 2023-01-24
Make spelling of "favorite" consistent

kalkwarf:
"Favorite" was spelled using both UK English and US English variations: 113 favourite
64 favorite

I made them consistent, using the US spelling.

I can change this to support the UK spelling if that's preferred, but my informal survey of existing words appears to support US spellings.


Dimillian:
This is good!

PR #329, opened 2023-01-24
Escape markdown markup characters on their way into HTMLToMarkdown

### LINKED ISSUE #328: Bug: Posts incorrectly render marked up by markdown
HTMLtoMarkdown only escapes asterisks when it returns its marked up text.

So a post that has \*this\* in it will render correctly, whereas a post that has \_this\_ will render the word in italics.

Also the relatively common ¯\\\_(ツ)\_/¯ loses an arm.

xurble:
To prevent them being rendered as styled text after they come out.

Fixes #328

### DIFF
--- a/Packages/Models/Sources/Models/Alias/HTMLString.swift
+++ b/Packages/Models/Sources/Models/Alias/HTMLString.swift
@@ -4,7 +4,7 @@ import SwiftSoup
 import SwiftUI
 
 public struct HTMLString: Decodable, Equatable, Hashable {
-  public let htmlValue: String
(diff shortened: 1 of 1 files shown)


Dimillian:
I had to revert this, it was preventing from unescaping some HTML characters.

xurble:
It's difficult, because the bug is actually in the library. 

Which characters caused the problem? Maybe there's a way to be smarter about it?  I didn't see any problems when I did it, but obviously I wasn't looking in the right places.

Dimillian:
the ' for example was showing the % encoding.

PR #341, opened 2023-01-24
Update Localizable.stringsdict

kum4423:
Translate "new post" into Japanese


Dimillian:
It's already in, as you sent it to me in DM. But for next update, feel free to do a PR! Thanks!

PR #345, opened 2023-01-24
Added marked text range support

h1d3mun3:
# Objective

When entering text in iOS, some languages need to take into account a range called 'markedTextRange'.
This `markedTextRange` represents the concept of "pre-finalized text", and speakers of languages that require "conversion" for language input need to consider this "pre-finalized text" state.
https://developer.apple.com/documentation/uikit/uitextinput/1614489-markedtextrange

For example, if a Japanese-speaking user tries to input text in your IceCube application, the following problem will occur if you do not take the markedTextRange into account.

https://user-images.githubusercontent.com/31161372/214290646-62e00e1b-13d6-45d4-bdfc-b5a94ef2e8fc.MP4

This user wants to input the word "あ" and have "words starting with あ" converted before inputting the text, but currently TextView cannot acquire the markedTextRange state, so the text finalization process will be executed.

Originally, the user expected the following state, in which conversion candidates are displayed at the top of the software keyboard.

https://user-images.githubusercontent.com/31161372/214290714-7771f111-6df9-4381-94a6-def4b2461d38.MP4

Since `markedTextRange is supported by the library in the following PR, we have added a consideration on the IceCube side in this PR.

https://github.com/Dimillian/TextView/pull/2

## Supplemental

To make sure this fix works, it is necessary to change the textView reference using the following Patch.
[patch.diff.zip](https://github.com/Dimillian/IceCubesApp/files/10489668/patch.diff.zip)


Dimillian:
Thanks a ton!

Issue #93, opened 2023-01-18
[Feature request] Prepend the language list with the frequently used ones or allow marking languages as preferred


jsynowiec:
This is a follow-up request to #76. It would be nice to have the frequently used, or starred, languages at the top of the language list. I imagine that the majority of users will use two, maybe three languages interchangeably, so having to search for it requires repetitive, unnecessary actions.

Dean151:
Last used languages are already live. Is this still to be left open?

alper:
Here's proof that this is done.
![IMG_45B3F34540B5-1](https://user-images.githubusercontent.com/72429/220448443-c9587ee1-0dd6-4c3b-a620-66b082f70079.jpeg)

PR #353, opened 2023-01-24
Prepend language list with recently used languages

scalbatty:
Adresses issue #93 

Rule: when the user explicitly selects a language when posting, when the post is successful, register the selected language as recently used. The recently used language array contains up to 3 distinct items, starting with the most recently used.

Simulator Screen Shot - iPhone 14 Pro - 2023-01-24 at 20 23 49

### DIFF
--- a/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
@@ -281,6 +281,7 @@
 "status.editor.drafts.navigation-title" = "Drafts";
 "status.editor.error.upload" = "Error uploading";
 "status.editor.language-select.navigation-title" = "Select Language";
+"status.editor.language-select.recently-used" = "Recently Used";
 "status.editor.media.edit-image" = "Edit Image";
 "status.editor.media.image-description" = "Image description";
 "status.editor.mode.edit" = "Editing your post";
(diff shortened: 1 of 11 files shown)


IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -281,6 +281,7 @@
 "status.editor.drafts.navigation-title" = "Drafts";
 "status.editor.error.upload" = "Error uploading";
 "status.editor.language-select.navigation-title" = "Select Language";
+"status.editor.language-select.recently-used" = "Recently Used";
```
EvilOne, line 284: Please copy this line to the other languages that the app provides. Then, the language updaters will see the English and create a PR to _fix_ it.

Thanks!

Dimillian: Yep, thanks! This need to be duplicated in every locale files or else the raw value  (key) will show.

scalbatty: I'm on it

Packages/Status/Sources/Status/Editor/Components/StatusEditorAccessoryView.swift
```diff
@@ -206,22 +225,43 @@ struct StatusEditorAccessoryView: View {
       .font(.scaledCallout)
   }
 
-  private var availableLanguages: [(String, String?, String?)] {
+  private struct Language: Identifiable, Equatable {
```
Dimillian, line 228: Much better :D

Dimillian:
All good, will merge as soon as the localizations files are updated.

PR #363, opened 2023-01-25
Akkoma/Pleroma support

FahimF:
* Support for Akkoma (and Pleroma) servers so that JSON parsing of objects does not fail/crash by ensuring that missing objects from Akkoma/Pleroma feeds are handled gracefully.
* Also fixes an issue with the Client which was sending POST data as URL parameters instead of in the request body. While this appears to work for Mastodon, it does break on other servers including Akkoma.


Packages/Models/Sources/Models/Account.swift
```diff
@@ -23,52 +23,99 @@ public struct Account: Decodable, Identifiable, Equatable, Hashable {
...
-  public let lastStatusAt: String?
-  public let fields: [Field]
-  public let locked: Bool
-  public let emojis: [Emoji]
-  public let url: URL?
-  public let source: Source?
-  public let bot: Bool
-  public let discoverable: Bool?
-
+  public var id = ""
```
Dimillian, line 26: Why can't it stay let? Even with a custom decoder there is no issue with the models being non initialised let or am I missing something?

FahimF: Just a matter of personal preference — I prefer to initialize variables where I can so that I can have an empty `initi` method ... or as close to empty as possible. But if you want to leave it as a `let` that should work too in cases where the value is guaranteed to be there in the received JSON. But if the value is not there in the received JSON for some reason, then you have to ensure that you explicitly set a value.

Dimillian:
Note sure the POST body works? Tried to post a post with special characters and I got a 422 empty status text.

Packages/Models/Sources/Models/Account.swift
Dimillian: But for example, at the API level, non logged statuses don't send some Bool value. I agree it looks weird but it was to stay as close as possible to the API output. And it looks ways cleaner too. Manual init are just for placeholder anyway the way I'm doing it.

Dimillian:
Up until now the some of the requests were sent in JSON, like status posting / editing. I was aiming at having everything sent in JSON in a near future. Does Plemora supports it?

Packages/Models/Sources/Models/Account.swift
FahimF: Those bools are now caught by the decode code since the value is filled in only if the value is there, otherwise it falls back on the default value set at init. The issue is some of the values which were supposed to be there for the Mastodon API are treated as optionals by some of the other APIs ...

Dimillian: I'm fine having more optionals let? if we can avoid manual decoding and CodingKeys. I didn't looked at the change in details yet at why we need custom decoders.

FahimF:
I had used the POST body code for a few weeks earlier, but I don't think I tried special characters. So can't tell you for sure but that's basically the same code (with some variation) that I use in multiple clients written for Swift code. Not saying that it doesn't have an error — I hesitate to definitively say something is (or is not) without testing ... But for general cases, it should work. Not sure about special characters though. The body might need to have that URL encoded and I might have forgotten to add that ...

Packages/Models/Sources/Models/Account.swift
FahimF: The custom decoder (in my case) was just so that I can catch decoding errors more easily. The `try` block lets me know what was found and what was not and so instead of fixing one variable at a time, testing again, and figuring out what crashes next, I just let the console logging tell me which variable caused and issue and can easily add  another `try` block for that ... For me, it makes debugging easier, and when you encounter a new system (besides Mastodon and Akkoma/Pleroma) it will make life easier for the next dev ...

May I ask why you prefer to avoid manual decoding?

Dimillian: Because it's more code to maintain. This project is about the less code, the better. Why would I want to write CodingKeys when I already have the variables as codingkeys?

FahimF: Totally your call. I have the code working the way I want at my end and was submitting a PR in case it helps you provide support for Akkoma and other instances. But if you want to approach it a different way, that's absolutely your prerogative 😄 

Please feel free to close the PR if it doesn't match up with your goals/aims. I am fine with that.

Dimillian: I totally want to support both instances kind, so thank you a lot for your work! I'll look into the specific a bit later and pick things from your PR. I want to stay close the code I have right now tho, yes :)

FahimF: Sounds good 😄 I'm bouncing back and forth between several different Mastodon clients at the moment and contributing where I can, but unfortunately, that means that I don't have a lot of time to spend on one particular client. But if there's some way I can help quickly, do let me know since the more clients we have for Mastodon, the better it is ... At least in my opinion 🙂

Dimillian:
Closing for now, will resume the work on this a bit later.

PR #366, opened 2023-01-25
Hotfix for SC localization

Jerry23011:
- Removed a repetition in the translations.
- Added new translations

What do you think @nixzhu :)


nixzhu:
LGTM

PR #370, opened 2023-01-25
Sort lists and tags in timeline filter button menus

DD-P:
Sorts alphabetically by `List.title` and `Tag.name` at the point of display in TimelineTab.

Alternatively they could be sorted in `CurrentAccount.fetchLists()` and `.fetchFollowedTags()`.


Dimillian:
That's nice!

Dimillian:
I wonder what is the default API sort? Added / created date?

DD-P:
I thought it might, but it doesn't seem to be. Nor id or hashValue as far as I can tell.  Looks like the API calls through to 

`perform_request_with_collection(:get, '/api/v1/lists', options, Mastodon::List)`

in mastodon-api/lib/mastodon/rest/lists.rb

PR #372, opened 2023-01-25
Updated NL localizations, part 5

icod:
- For content settings section

@Dimillian, I noticed that the builds get these red crosses after my PRs are merged into `main`. What can I do to prevent this?


Dimillian:
@icod Don't worry it's because if a deploy is in progress and I merge something, the previous build get cancelled :) It's excepted.

PR #380, opened 2023-01-25
fix: Update some german translations

LekoArts:
Hello!

First PR 🎉 
I started using the app and found some possible improvements to the german translation :)

### DIFF
--- a/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
@@ -11,7 +11,7 @@
 "placeholder.loading.long" = "Laden, laden, laden ....";
 "placeholder.loading.short" = "Laden ...";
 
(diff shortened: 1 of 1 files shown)


IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -55,14 +55,14 @@
 "settings.display.navigation-title" = "Anzeigeeinstellungen";
 "settings.display.restore" = "Standard wiederherstellen";
 "settings.display.section.display" = "Anzeigen";
-"settings.display.section.theme" = "Thema";
+"settings.display.section.theme" = "Design";
```
LekoArts, line 58: While "Thema" is already a german word, I checked what other apps do (e.g. Notion, Things) and they often use "Design" which is acceptable in german and IMO fits better

PR #382, opened 2023-01-25
Move notification and reboosts actions into account menu

jdanthinne:
Not sure about this one, but as some users don't understand the correct state of the notify and reboosts button (is it on, is it off?), this PR moves those 2 actions into the account menu, with clear action texts (Enable/Disable Notifications, Show/Hide Boosts).

Simulator Screen Shot - iPhone 14 Pro - 2023-01-25 at 16 13 50


jdanthinne:
(and made the menu label icon .circle.filled as it is sometimes very hard to see on some backgrounds)

Dimillian:
Can you also leave both button? I think have both entries is perfectly fine. I love having them next to the follow button + in the menu. should be self explanatory enough with the matching icons.

jdanthinne:
I've put back the two buttons.

PR #402, opened 2023-01-26
Allow editing photo description/alt-text after posting

### LINKED ISSUE #401: Feature request: editing a post with an image should allow adding/editing the photos description/alt-text
Given: You've already created a post with an image included, and without any description/alt-text for the image
When: The user edits the post
Then: The user should be able to update the description/alt-text of the image

There is some code here that prevents the add/edit description menu item from showing, which is intended due to how the Mastodon API works I assume.

https://github.com/Dimillian/IceCubesApp/blob/main/Packages/Status/Sources/Status/Editor/Components/StatusEditorMediaView.swift#L124

```
if !viewModel.mode.isEditing {
      Button {
        editingContainer = container
      } label: {
        Label(container.mediaAttachment?.description?.isEmpty == false ?
          "status.editor.description.edit" : "status.editor.description.add",
          systemImage: "pencil.line")
      }
    }
```

Removing the if block will allow the following workflow (albeit with some rough edges):

1. Select the image and delete
2. Add the image back
3. Select the image and be able to add a description

If the user tries to Add/edit the description without deleting the original image first, the change will not go through. Perhaps we could programmatically delete the image first and then re-add prior to displaying the modal? Maybe there is a way to check if the original image was removed, and not display the add/edit description menu item in that case?

I know this is not the best workflow, but at least it gives the user an option to add descriptive text to the image, without having to delete the whole post, and start from scratch.

joebarbere:
Naively fixes #401
I'll do some more research to see if we can hide the add description menu item based on whether the image has been uploaded.

### DIFF
--- a/Packages/Status/Sources/Status/Editor/Components/StatusEditorMediaView.swift
+++ b/Packages/Status/Sources/Status/Editor/Components/StatusEditorMediaView.swift
@@ -121,14 +121,12 @@ struct StatusEditorMediaView: View {
 
   @ViewBuilder
   private func makeImageMenu(container: StatusEditorMediaContainer) -> some View {
-    if !viewModel.mode.isEditing {
-      Button {
+    Button {
         editingContainer = container
-      } label: {
-        Label(container.mediaAttachment?.description?.isEmpty == false ?
+    } label: {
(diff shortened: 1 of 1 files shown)


Dimillian:
Closing for now, this is not something we can do properly.

PR #404, opened 2023-01-26
Add link to system settings

kfdm:
I've often found it useful when an app had a shortcut directly to the system settings. Is this something you might consider?

It's unstyled/untranslated right now because I just wanted a quick sketch to see if this was something you would be interested in, before I actually dug into the translated strings :) 

system settings
ios settings


Dimillian:
Super useful but could you add the `settings.system` to all localizable files? and add the correct localizations for the English one? (you can leave english one for all others too)

kfdm:
No problem at all! I wanted to make sure this was something that seemed reasonable before I tried to figure out the language strings :)

kfdm:
- Rebased against most recent main
- Added Localizable.strings

Let me know if there's anything else I should do :)

Dimillian:
Looking good thanks! I might need to review the tint.

PR #407, opened 2023-01-26
Suggestion: Improve Status text visibility when context menu is opened.

h1d3mun3:
## Motivation

I'd like to improve text readability and visibility when status context menu is opened in Light mode settings.

## Changed 

- make it set background color to S tatusRowView's VStack which use `theme.primaryBackgroundColor`

### Light
before | after
--- | ---
Light-Before mp4 | Light-After mp4

### Dark

before | after
--- | ---
Dark-Beforee mp4 | Dark-After mp4

## Consideration

If non color specified to StatusRowView's VStack, make it easy color management on view which useing StatusRowView.

So, if you intended this, please close this pull-request.

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusRowView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowView.swift
@@ -62,6 +62,7 @@ public struct StatusRowView: View {
           }
         }
       }
+      .background(theme.primaryBackgroundColor)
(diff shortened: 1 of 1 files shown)


Dimillian:
I'll close for now because it creates other issues I need to fix first. But thanks, the no background is definitely not intended!

PR #432, opened 2023-01-27
Spanish update 27-01

WedgeSparda:
- Updated spanish localization with new keys
- Added missing NSLocalizedString for filters
- Removed duplicated localization key after merge


Dimillian:
Some problem with this PR, could you take a new main? Thanks!

PR #436, opened 2023-01-27
Update Simplified Chinese Localization

Jerry23011:
I think this will be a rather quick review, @nixzhu :)
The strings cater to the translations of Mastodon


nixzhu:
LGTM

Jerry23011:
刚刚去官方看了一下，`account.detail.about` 叫“附加信息”

PR #445, opened 2023-01-27
Add plurals for Simplified Chinese 

Jerry23011:
Hope this is fast enough to be loaded on the next release


Jerry23011:
It's a quite late here 💤
I'll just leave the message here @nixzhu

Dimillian:
No worries it's all good :)

PR #450, opened 2023-01-27
Spanish update 27-01

WedgeSparda:
- Updated spanish localization
- Fixed missing NSLocalizedString


IceCubesApp/Resources/Localization/it.lproj/Localizable.strings
```diff
@@ -201,8 +201,12 @@
 "account.posts" = "Post";
 "account.relation.follows-you" = "Ti segue";
 "account.joined" = "Si è unito il giorno";
+<<<<<<< HEAD
```
Dimillian, line 204: Conflict :)

WedgeSparda: Fixed and commited. Thanks for catching it.

PR #452, opened 2023-01-27
Add swipe-to-dismiss to QLPreviewController

gregggreg:
Work-around issue with QLPreviewController not supporting swipe-to-dismiss and pinch-to-dismiss when presented from SwiftUI by creating a transparent UIViewController wrapper around QLPreviewController that presents it using UIKit instead.

I tried many different ways to workaround this issue and this was the best way that I could figure it out. I've also filed feedback to Apple about the fact that SwiftUI has different behavior when presenting QLPreviewController than UIKit, so if they ever fix it, this workaround would be no longer needed.


Dimillian:
I can't believe you did that, this is glorious!

Dimillian:
It's working good even on macOS!

PR #448, opened 2023-01-27
Add Brazilian Portuguese Localization

arthurgivigir:
Hey from a new Mastodon Fan!
This is my first contribution on this amazing app, adding Brazilian Portuguese 🇧🇷.
Any problem, please reach me out!

PR #453, opened 2023-01-27
Polish localization

AndrzejRozga:
Polish localization by AndrzejRozga


Jerry23011:
Hi, I think you need to do this in Xcode? It seems that some files are missing (correct me if I'm wrong, though)
Please see #448

Dimillian:
Thanks, yes @Jerry23011, I'm merging and I'll properly add it in the next commit.

PR #454, opened 2023-01-27
Russian localization

FedorSit:
Hello, I'm a newcomer on GitHub, so I can make mistakes, I made as my contribution to this fantastic project, Russian localization)


Jerry23011:
Hi, I think you need to do this in Xcode? It seems that some files are missing (correct me if I'm wrong, though)
Please see #448

PR #455, opened 2023-01-27
Add indicator to which language a status is set

vollkorntomate:
I added a small indicator showing which language was set for a post, similar to Mastodon's web UI. Unfortunately, this often doesn't match the actual content of a post, but that's a different problem.

image

Since this feature is probably only relevant to a few people (or power users), I could add a setting to only show if the user enables it (default: not visible). Please let me know if such a setting would make sense.

### DIFF
--- a/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
@@ -289,6 +289,7 @@
 
 // MARK: Package: Status
 "status.action.translate" = "Übersetzen";
+"status.action.translate-from-%@" = "Übersetzen aus %@";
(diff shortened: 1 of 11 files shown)


Dimillian:
Yes, this line hardly go on one line already, a settings near the translate button with a default to off would be nice thanks.

vollkorntomate:
While thinking about this a little further, I came up with a slightly different (and in my opinion better) solution: the "translate" button now shows from which language the post will be translated.

Here's an example ("Translate from Japanese"):
Bildschirm­foto 2023-01-29 um 17 30 22

Bildschirm­foto 2023-01-29 um 17 30 39

That makes it clear, in what language a post was created (or at least which language is set for that post) and is even more descriptive than two small letters somewhere.

Dimillian:
Very smart!

PR #456, opened 2023-01-28
Fixed modal showing up everytime. 

PrajeetShrestha:
When unauthenticated, everytime user switched tab, the add account modal is showing up now. Added a code so that the popup shows only one time after user initiates the app.

### DIFF
--- a/IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
+++ b/IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
@@ -39,6 +39,9 @@ struct TimelineTab: View {
         .id(currentAccount.account?.id)
     }
     .onAppear {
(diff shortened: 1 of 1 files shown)


Dimillian:
I'll close this one as showing the modal when not logged often is actually on purpose.

PrajeetShrestha:
Should the popup be shown every time user switched to timeline view from settings? It will be inconvenient for people who want to explore the app without logging in in my opinion.

Dimillian:
I had enough problem during App Review to change this behaviour now :/ I know it's less than stellar but at this point I'm too afraid to change it for them lol

PrajeetShrestha:
I can only imagine your frustation with appstore haha. Just for your information though, the same behavior is not present in iPadOS / macOS because the onAppear for timeline will only be called once in those platform.

PR #461, opened 2023-01-28
Fix: consolidated notifications

jdanthinne:
• Correctly handle notifications paging (was not using the correct last notification id)
• Don't consolidate followers when selecting only Followers as Notifications filter


Dimillian:
Perfect!

PR #462, opened 2023-01-28
This is a more conservative fix for #328

xurble:
Having checked the markdown -> attributedtext conversion, lots of things I was escaping before are actually ingored.

This change only escapes things that are definitely rendered and as far as I can tell doesn't break anything now.

Test post to look at with maximal Markdown: https://mas.to/@elbrux/109743775703438333

My only worry is that it makes things slower, but I did some profiling and it didn't show anything.  So that could just be running under the debugger, it could be paranoia or it could be my lack of skill with the profiler.


Dimillian:
This is only done at decoding time, so it's fine

Dimillian:
Seems all good! Thanks

PR #472, opened 2023-01-28

Jerry23011:
### Changes
- add new translations

Part of the translations in "Filters" are from Mastodon
Could you take a look, @nixzhu  :)


nixzhu:
LGTM

PR #476, opened 2023-01-28
Fix french localizations

jdanthinne:


Packages/Models/Sources/Models/Alias/ServerDate.swift
```diff
@@ -13,7 +13,7 @@ extension ServerDate {
 
   private static var createdAtRelativeFormatter: RelativeDateTimeFormatter = {
     let dateFormatter = RelativeDateTimeFormatter()
-    dateFormatter.unitsStyle = .abbreviated
```
Dimillian, line 16: Are you sure about that?

Dimillian:
Ah Nice!

Packages/Models/Sources/Models/Alias/ServerDate.swift
jdanthinne: Yes, tested in French and English. English is same a before, and French went from "-1h." to "il y a 1h".

PR #502, opened 2023-01-28
Fix for launching app using just URL scheme

gregggreg:
Skip blank, host-less URLs so that app launchers can launch the app using just the scheme (e.g. icecubesapp://) without invoking a safari error page


gregggreg:
To test with current version, open Safari and type in: icecubesapp:// and see what happens.

Dimillian:
Thanks!

PR #503, opened 2023-01-29
Retain original ordering of media URLs

gregggreg:
For posts with multiple images, the images loaded in the media viewer lost the original ordering from the post. This re-sorts the URLs by their original order.


Dimillian:
Nice!

PR #508, opened 2023-01-29
Fix Theme not overriding UserInterfaceStyle

divadretlaw:
When adding the option to follow the system theme, the code to override the system theme was removed, but when the system theme is not overridden and for example a Dark Theme is selected while being in light mode, then a lot of UI elements would be light mode, which looks out of place.

This restores the old behavior where the theme overrides the UserInterfaceStyle on all windows while still respecting the "Follow System Theme" option

### DIFF
--- a/Packages/DesignSystem/Sources/DesignSystem/ThemeApplier.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/ThemeApplier.swift
@@ -38,6 +38,7 @@ struct ThemeApplier: ViewModifier {
           theme.selectedSet = colorScheme == .dark ? sets.dark.name : sets.light.name
         }
         setWindowTint(theme.tintColor)
+        setWindowUserInterfaceStyle(from: theme.selectedScheme)
         setBarsColor(theme.primaryBackgroundColor)
       }
       .onChange(of: theme.tintColor) { newValue in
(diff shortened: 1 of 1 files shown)


Dimillian:
Are you sure of that? I've removed it because it was setting a custom scheme preventing SwiftUI from receiving the system scheme.

Packages/DesignSystem/Sources/DesignSystem/ThemeApplier.swift
```diff
@@ -58,6 +62,26 @@ struct ThemeApplier: ViewModifier {
   }
 
   #if canImport(UIKit)
+    private func setWindowUserInterfaceStyle(from colorScheme: ColorScheme) {
+      guard !theme.followSystemColorScheme else {
+          setWindowUserInterfaceStyle(.unspecified)
+          return
```
Dimillian, line 68: That's fixing the problem in my comment right?

divadretlaw: Yes, this undoes the overriding of the user interface style by resetting it to `.unspecified` in case the system theme is followed

A recording of the issue when not following system mode without this PR.

iPhone 14, iOS 16.2 - Light theme for system

| Settings | Issue |
|:-:|:-:|
| Settings | https://user-images.githubusercontent.com/6899256/215322215-99f69d27-21eb-4d45-83c2-399c55fea5fa.mov |

PR #518, opened 2023-01-29
Improved stretchy account header

mbuchetics:
This improves the behaviour of the sticky & stretchy header on the account detail screen. Previously pulling down resulted in the content being scrolled too fast and there was a gap between header and content. This was caused by modifying the header height which had an impact on the entire scroll view. By doing putting the header in an overlay (and keeping the rectangle above the content at a fixed height), it feels much better with a more natural resistance.

I also changed some constants to show the navigation bar title earlier and also optimized the transitioning between filled and regular ellipsis icon in the navigation bar.

https://user-images.githubusercontent.com/211709/215345560-4749430a-e659-4801-aa22-6785aee3abd5.mp4


Dimillian:
That soooo nice. Will wait a bit before merging as I have some trouble with the CI.

PR #519, opened 2023-01-29
Add the ability to set a custom font

PaulDoesDev:
Note: The FontPicker as a modal feels a little strange on Catalyst, but when I tried presenting it as a popover instead it seems to anchor to the top left, regardless of what I did. I felt this was the better of the two options.


Dimillian:
This is awesome and now it's even easy to add custom font scaling on iOS too

PR #524, opened 2023-01-29
Boost deduplicating

xurble:
This will prevent boosts from being shown over and over.  Every boost seen is cached, and only the first boost you saw is shown therafter.

Only works on the main timeline at the moment.  Cache size is limited to 100 items at the moment, but it could easily be increased.  

Since I started this, Ivory has announced this feature.  They don't have an option to turn it on/off so I haven't done it either. Mastodon actually already suppresses dupe boosts, but only within the same API call so it seems like it's the correct behaviour.


Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -26,7 +29,9 @@ public class StatusRowViewModel: ObservableObject {
 
   @Published var translation: String?
   @Published var isLoadingTranslation: Bool = false
-
+  
+  @Published var seen = false
```
Dimillian, line 33: It doens't need to be published since it have no impact on the UI.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -62,6 +67,14 @@ public class StatusRowViewModel: ObservableObject {
 
     isFiltered = filter != nil
   }
+    
+  func markSeen() {
+    // called in on appear so we can cache that the status has been seen.
+    if !seen {
+      ReblogCache.shared.cache(status, seen: true)
```
Dimillian, line 74: You should wrap that in a background DispatchQueue, that could definitely hang the UI.

Dimillian:
I still think an option to turn it off would be nice, I can add it myself later tho. Really need to test performances for the seen part on this one.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
Dimillian: Should you cache only if this a reblog?

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
xurble: Yep.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
xurble: The cache tests if it's a reblog so it does get checked, can move the check out here if you prefer.

xurble:
The cache promises constant time access and didn't seem like it was struggling in my testing.  I wasn't sure how dispatching a bunch of threads would compare.  I only have my 1 phone to test on which is a 13 Pro.  Maybe older phones would stuggle with one approach v the other?

That whole seen thing will become moot anyway if you move to lists because the onappear would fire as soon as the cell was created and lists aren't lazy are they?

So I'm happy to be guided by you.  Stick to as is, change to threads or just rip out seen altogether, it's arguably overkill.

xurble:
Sorry, I forgot about the setting, I can do it.  

It's made me think though.  This will only cache after first seen boost, but there's a more aggressive implementation where it hides all boost of posts made by accounts you follow on the assumption that you probably saw it when it was first posted.  I didn't implement that by default  but as an option I think it makes sense?

Are you good with me putting that in?

Dimillian:
> The cache promises constant time access and didn't seem like it was struggling in my testing. I wasn't sure how dispatching a bunch of threads would compare. I only have my 1 phone to test on which is a 13 Pro. Maybe older phones would stuggle with one approach v the other?
> 
> That whole seen thing will become moot anyway if you move to lists because the onappear would fire as soon as the cell was created and lists aren't lazy are they?
> 
> So I'm happy to be guided by you. Stick to as is, change to threads or just rip out seen altogether, it's arguably overkill.

I didn't saw it was just a dict access and that the disk save is already done every 30 secs on a background queue. So sound fine.

Dimillian:
> Sorry, I forgot about the setting, I can do it.
> 
> It's made me think though. This will only cache after first seen boost, but there's a more aggressive implementation where it hides all boost of posts made by accounts you follow on the assumption that you probably saw it when it was first posted. I didn't implement that by default but as an option I think it makes sense?
> 
> Are you good with me putting that in?

I think this is even better and simpler? But yeas definitely need the settings in this case.

xurble:
I didn't realise that followers weren't always stored on an account.  They're loaded when you click on the followers count on your profile.  Suppressing boosts from followers would required loading and caching followers at startup, which I am happy to do, but it's more of a change so I thought I'd check you actually want that first.

Dimillian:
Not sure we want to go in this direction. It feels quite heavy compared to the philosophie of the app. I do want to have local storage and caching but I'm not there yet. I've explored a few solutions this weekend for the timeline.

xurble:
I'll leave it for now. I'll just make the setting cache on/off with the current behaviour. 

If you ever start local caching of followers, I can revisit it

xurble:
I think this should be done now.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -62,6 +67,14 @@ public class StatusRowViewModel: ObservableObject {
 
     isFiltered = filter != nil
   }
+    
+  func markSeen() {
```
Dimillian, line 71: What about also having the settings check here? So it's not adding anything to the cache if the feature is off?

Packages/Env/Sources/Env/UserPreferences.swift
```diff
@@ -28,6 +28,8 @@ public class UserPreferences: ObservableObject {
   @AppStorage("autoplay_video") public var autoPlayVideo = true
   @AppStorage("chosen_font") public private(set) var chosenFontData: Data?
 
+  @AppStorage("suppress_dupe_reblogs") public var suppressDupeReblogs: Bool = true
```
Dimillian, line 31: This should be off by default, to honour the original feed.

Dimillian:
Two more comments and then all good to merge 🚀

xurble:
I think that's got it now.

sothawo:
Just installed it locally on my iPhone, translations seem to be missing.

image

Dimillian:
Fixed

mike3k:
Can we increase the reblog cache size to 200? 100 is too small when you follow a lot of people and I still end up seeing many duplicated boosts. Ivory seems to be more aggressive at hiding them.

xurble:
That would almost certainly be fine, probably even more.  I just started conservatively.

mike3k:
> That would almost certainly be fine, probably even more. I just started conservatively.

I just did a tiny PR for it, increasing it to 300.

Issue #127, opened 2023-01-19
Improve profiles without header images


AndrewHaglund:
If a profile doesn't have a header image, the header image area is grey (expected: transparent?) and tapping opens an empty center modal (and on the Mac, an empty Quick Look window also opens).

![Screen 2023-01-19 at 3 50 20 PM](https://user-images.githubusercontent.com/5075921/213569835-349e0a5c-bd65-402c-8340-399a81eb68cd.jpg)

![Screen 2023-01-19 at 3 48 31 PM](https://user-images.githubusercontent.com/5075921/213569717-b6a68559-3c0a-4906-9ac6-5e36d701c027.jpg)

PR #526, opened 2023-01-29
Better profile header

Dean151:
Motivation:
- When a header is missing, we can still open quicklook #127
- When a header is missing, we're using a basic gray background

Proposition:
- Stop opening header & avatar quicklook when it's missing
- Using secondary background instead of standard gray when header is missing

Before / After:
before after

### DIFF
--- a/Packages/Models/Sources/Models/Account.swift
+++ b/Packages/Models/Sources/Models/Account.swift
@@ -43,6 +43,14 @@ public struct Account: Decodable, Identifiable, Equatable, Hashable {
   public let bot: Bool
   public let discoverable: Bool?
 
+  public var haveAvatar: Bool {
+    return header.lastPathComponent != "missing.png"
+  }
+
+  public var haveHeader: Bool {
+    return header.lastPathComponent != "missing.png"
+  }
+
   public static func placeholder() -> Account {
     .init(id: UUID().uuidString,
           username: "Username",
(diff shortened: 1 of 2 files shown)


Packages/Models/Sources/Models/Account.swift
```diff
@@ -43,6 +43,14 @@ public struct Account: Decodable, Identifiable, Equatable, Hashable {
   public let bot: Bool
   public let discoverable: Bool?
 
+  public var haveAvatar: Bool {
+    return header.lastPathComponent != "missing.png"
```
melwaraki, line 47: I believe this should be `avatar.lastPathComponent`?

Dimillian: ✅

Packages/Models/Sources/Models/Account.swift
```diff
@@ -44,7 +44,7 @@ public struct Account: Decodable, Identifiable, Equatable, Hashable {
   public let discoverable: Bool?
 
   public var haveAvatar: Bool {
-    return header.lastPathComponent != "missing.png"
+    return avatar.lastPathComponent != "missing.png"
```
Dean151, line 47: Nicely spotted. ⌘+C is never a good thing after midnight

Dean151, line 47: @Dimillian Nicely spotted. ⌘+C is never a good thing after midnight

PR #549, opened 2023-01-31
Highlight timeline row where I am mentioned

EvilOne:
Short PR to highlight a timeline row where the current user is mentioned.
CleanShot 2023-01-30 at 18 24 58@2x

### DIFF
--- a/Packages/Status/Sources/Status/List/StatusesListView.swift
+++ b/Packages/Status/Sources/Status/List/StatusesListView.swift
@@ -1,10 +1,12 @@
 import DesignSystem
+import Env
 import Models
 import Shimmer
 import SwiftUI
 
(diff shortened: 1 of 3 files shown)


Dimillian:
I'm careful with this one, will need to do some checks when I'll have time but I think it could create some performances issues, as the var is computed at scroll time.

seanlifetime:
Sounds like a plan. Performance is important. 

Thanks!

Dimillian:
Closing it for now, I still want it in, but I'll think of a different way to do it. I'm super careful now when adding computed stuff to the timeline.

EvilOne:
I am working on something that doesn't do it at the var level for the timeline.

PR #551, opened 2023-01-31
make share and copy links translatable

maco:
I was filing a ticket about the following when I realized I could just fix it, so here's an explanation of the bug, along with a PR for the fix. I'm unsure if I need to add the untranslated English one to every other language file, or if it'll automatically fall back to English. (I would guess it varies by programming language & framework.) So, if I need to add the  English as placeholders in other files, please just let me know.

The `ShareLink()` in `AccountDetailView.swift` (and elsewhere) is called without passing an argument for the label, so the text is left untranslated. For example, here's a screenshot from my phone, which is in Spanish:
IMG_BF2AEBBAB7B2-1

Similarly, the `ShareLink` in `StatusCardView.swift` _does_ have a label, but it's a hard-coded string, not translatable. Same for the copy option.
IMG_29F5617133DC-1

### DIFF
--- a/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
@@ -185,6 +185,7 @@
 "account.action.unblock" = "Unblock";
 "account.action.mute" = "Mute";
 "account.action.unmute" = "Unmute";
+"account.action.share" = "Share this account";
 "account.boosted-by" = "Boosted by";
 "account.detail.about" = "About";
 "account.detail.familiar-followers" = "Also followed by";
@@ -330,6 +331,8 @@
 "status.action.unfavorite" = "Unfavorite";
 "status.action.unpin" = "Unpin";
 "status.action.view-in-browser" = "View in Browser";
+"status.card.share" = "Share this link";
+"status.card.copy" = "Copy this link";
 "status.draft.delete" = "Delete Draft";
 "status.draft.save" = "Save Draft";
 "status.editor.ai-prompt.correct" = "Correct text";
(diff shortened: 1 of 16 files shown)


IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -169,6 +169,7 @@
 "account.action.unblock" = "Unblock";
 "account.action.mute" = "Mute";
 "account.action.unmute" = "Unmute";
+"account.action.share" = "Share this account";
```
Dimillian, line 172: All the new key need to be added to every files. You can add the english translation as a placeholder.

Dimillian:
Thanks!

IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
Dimillian: Hmm it still not on every localizable files

maco: Are there ones that aren't named Localizable.strings that I should be looking for? Because

```
❯ find . -name Localizable.strings | wc -l
      14
```

Says there are 14 of them total. And these say I have inserted those placeholders into 14 different files of that name:

```

IceCubesApp on  make_shares_translatable [⇡]
❯ find . -name Localizable.strings -exec grep "account.action.share" {} \; | wc -l
      14

IceCubesApp on  make_shares_translatable [⇡]
❯ find . -name Localizable.strings -exec grep "status.card.share" {} \; | wc -l
      14

IceCubesApp on  make_shares_translatable [⇡]
❯ find . -name Localizable.strings -exec grep "status.card.copy" {} \; | wc -l
      14
```

So I don't understand what I missed.

maco: Oh 🤦‍♀️ I missed `git push`

Dimillian: I guess you'll have quite a bit of conflict now

PR #563, opened 2023-01-31
Consolidate new notifications with the latest one

jdanthinne:
This remove the need of a shadow array of notifications, and try to consolidate the newly received notification with the latest notifications, if of the same type.
Made some cleanup in the Models as well not to pollute the NotificationsViewModel.


Dimillian:
Very nice!

PR #573, opened 2023-01-31
[proposal] Respect OS localization to define translation

jairhenrique:
The translation button is only visible for toot's that are in my operating system language. I think it makes more sense to respect the system configuration and present the button for all other languages.

### DIFF
--- a/IceCubesApp/Resources/Localization/ca.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/ca.lproj/Localizable.strings
@@ -290,7 +290,7 @@
 
 // MARK: Package: Status
 "status.action.translate" = "Tradueix";
-"status.action.translate-from-%@" = "Translate from %@";
+"status.action.translate-to-%@" = "Translate to %@";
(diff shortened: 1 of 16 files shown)


Jerry23011:
Hi, I contribute to the localization, could you give an example to show what is the difference between "from" and "to"? Thanks :)

Dimillian:
Not sure about this one at all. I can comment on English and French, this make sense in current localization, you're translating from the post language. Even if it's not visually what you see (nothing prevent the user to set the post in English when their posting in another language). Closing this is for now, but don't hesitate to add more comments.

jairhenrique:
@Jerry23011 @Dimillian  the main language is not necessarily English. My main language is Brazilian Portuguese (pt-BR) and the translation button appears only for toots in (pt-BR) to be translated to (en).

For me it is more relevant to translate anything different from (pt-BR) **to** (pt-BR) and not **from** (pt-BR) to (en). So I considered reversing the language assignment order to what is native to the OS and thus showing the translation button for that language.

<div style="float: left">



</div>

PR #579, opened 2023-02-01
Improved new posts view with avatar of the posters

PrajeetShrestha:
Added a expandable pending notification indicator with avatars of the accounts who posted the status. 

Note: 

- Code to scroll to last pending status is not altered  in this pull request, and does not seem to be at working state as of now. 
- There are two new strings added: expanded and shrinked. I have added it to english string resource Localizable(english) only.


PrajeetShrestha:
Looks like there were some changes in codebase. I will reopen after verifying my changes works as expected.

Packages/Timeline/Sources/Timeline/PendingStatusesObserver.swift
```diff
@@ -1,51 +1,86 @@
...
+import DesignSystem
+import Combine
 
 @MainActor
 class PendingStatusesObserver: ObservableObject {
-  let feedbackGenerator = UIImpactFeedbackGenerator(style: .light)
+  @AppStorage("expanded") public var isExpanded: Bool = false
   
-  @Published var pendingStatusesCount: Int = 0
+  @Published var pendingStatuses: [Status] = []
```
Dimillian, line 11: Note sure we should store the whole status here. I've made it an hint so the diffing of SwiftUI could be very fast. It was slow as yous scrolled up prior to that. 
I'm ok to have a shadow non published variable holding status tho. And actually you had it, `pendingStatuses`

Packages/Timeline/Sources/Timeline/TimelineView.swift
```diff
@@ -53,9 +53,9 @@ public struct TimelineView: View {
         .listStyle(.plain)
         .scrollContentBackground(.hidden)
         .background(theme.primaryBackgroundColor)
-        if viewModel.pendingStatusesEnabled {
+//        if viewModel.pendingStatusesEnabled {
```
Dimillian, line 56: Still good to have this check.

PrajeetShrestha:
Tried other way to achieve this but I think unpublished status in "PendingStatusesObserver" would be more cleaner approach to do it.

Packages/Timeline/Sources/Timeline/TimelineView.swift
PrajeetShrestha: Removed. commented by mistake.

Packages/Timeline/Sources/Timeline/PendingStatusesObserver.swift
PrajeetShrestha: Made pendingStatuses , none published now

Dimillian:
Closing it for now, I like the idea but I'll rework it into something a bit different when I'll have time.

PR #580, opened 2023-02-01
Timeline: restore the selected timeline at the next startup

nekolaboratory:
Closes: #581

### DIFF
--- a/IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
+++ b/IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
@@ -20,7 +20,7 @@ struct TimelineTab: View {
   @State private var timeline: TimelineFilter
   @State private var scrollToTopSignal: Int = 0
   
-  @AppStorage("timelineFilter") public var timelineFilter: String = TimelineFilter.home.title
+  @AppStorage("last_timeline_filter") public var timelineFilter: String = TimelineFilter.home.title
 
   private let canFilterTimeline: Bool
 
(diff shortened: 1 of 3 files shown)


IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
```diff
@@ -19,6 +19,8 @@ struct TimelineTab: View {
   @State private var didAppear: Bool = false
   @State private var timeline: TimelineFilter
   @State private var scrollToTopSignal: Int = 0
+  
+  @AppStorage("timelineFilter") public var timelineFilter: String = TimelineFilter.home.title
```
Dimillian, line 23: should be named something like `last_timeline_filter`, same for the variable name, `lastTimelineFilter` right now it could be a bit confusing with the actual filter enum.

Dimillian:
I understand the feature request, but I might do it in a more safe way. Right now encoding the title could lead to many problem down the line. Better make the enum itself `Codable` so we could encode the whole filter and just re init it. RawRepresentable probably need to be extended to fit an enum or a `Codable` in it.

nekolaboratory:
Thank you for your advice.
I agree with your opinion and will revise it later.

Dimillian:
Closing it for now, I've made the timeline filter even more complexe, and this create a lot of side effects / issues. Will think about how to save / restore it correctly.

nekolaboratory:
@Dimillian After you closed this PR, I got the latest version, merged it, and checked it, but there was no particular problem.
Could you please reconsider merging?

Issue #568, opened 2023-01-31
Bug: View in Browser from a post just reloads the post in the app


lizadaly:
In version v.1.2.2 on iOS:

Selecting "View in Browser" from a post (either in the timeline or with the post opened in detail) only loads the post in the app, as if it had been clicked on from the timeline. This is true regardless of whether the browser setting is System or In-App.

I've definitely used this feature before, so it seems like it's a regression. I've tried force-quitting and restarting the app.

PR #585, opened 2023-02-01
More fixes for #568 - Open Url in browser

Dean151:
I saw the fix for StatusRowContextMenu, so this is a complement for this same bug accros the app

This is related with the changes we made to the router to "Open in Ice Cubes" 😬🫣

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusCardView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusCardView.swift
@@ -64,7 +64,7 @@ public struct StatusCardView: View {
         ShareLink(item: url) {
           Label("Share this link", systemImage: "square.and.arrow.up")
         }
-        Button { UIApplication.shared.open(url) } label: {
+        Button { openURL(url) } label: {
           Label("status.action.view-in-browser", systemImage: "safari")
         }
         Divider()
--- a/Packages/Status/Sources/Status/Row/StatusCardView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusCardView.swift
@@ -64,7 +64,7 @@ public struct StatusCardView: View {
         ShareLink(item: url) {
           Label("Share this link", systemImage: "square.and.arrow.up")
         }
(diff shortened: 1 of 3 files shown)


Packages/Status/Sources/Status/Row/StatusCardView.swift
```diff
@@ -64,7 +64,7 @@ public struct StatusCardView: View {
         ShareLink(item: url) {
           Label("Share this link", systemImage: "square.and.arrow.up")
         }
-        Button { openURL(url) } label: {
+        Button { UIApplication.shared.open(url) } label: {
```
Dimillian, line 67: But for the card, you still want to use the routeur right? It should respect the open in app browser vs open in system browser.

Packages/Status/Sources/Status/Row/StatusRowContextMenu.swift
```diff
@@ -8,8 +8,6 @@ struct StatusRowContextMenu: View {
   @EnvironmentObject private var currentInstance: CurrentInstance
   @EnvironmentObject private var routerPath: RouterPath
 
-  @Environment(\.openURL) var openURL
```
Dimillian, line 11: good cleanup.

Packages/Status/Sources/Status/Row/StatusCardView.swift
Dean151: Not sure where the card is located, I was following the "Open in browser" translation.
Problem is that currently, openURL called on a Mastodon url somehow always end up to be an in-app trigger from my understanding

Dimillian: It's fine for card, card are the big link thing in some statuses. Basically it'll route in the app if it's a mastodon link and route to the correct browser if it's not. So it's fine to leave untouched there.

Dean151: Let it be just a *very* small cleanup PR then

PR #587, opened 2023-02-01
Highlight me on timeline -- better

EvilOne:
A better way to highlight _me_ on the timeline.

Comments are welcome. I want to get the idea out there as I look to clean things up a bit more.

### DIFF
--- a/Packages/Models/Sources/Models/Status.swift
+++ b/Packages/Models/Sources/Models/Status.swift
@@ -59,6 +59,8 @@ public struct Status: AnyStatus, Decodable, Identifiable, Equatable, Hashable {
     id + createdAt + (editedAt ?? "")
   }
 
+  public var shouldHighlight: Bool?
+
   public static func == (lhs: Status, rhs: Status) -> Bool {
     lhs.id == rhs.id
   }
--- a/Packages/Timeline/Sources/Timeline/TimelineView.swift
+++ b/Packages/Timeline/Sources/Timeline/TimelineView.swift
@@ -75,6 +75,7 @@ public struct TimelineView: View {
       if viewModel.client == nil {
         viewModel.client = client
         viewModel.timeline = timeline
+        viewModel.account = account
       } else {
         Task {
           await viewModel.fetchStatuses()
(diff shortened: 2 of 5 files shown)


Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
```diff
@@ -152,6 +154,15 @@ extension TimelineViewModel: StatusesFetcher {
                                                                   maxId: nil,
                                                                   minId: nil,
                                                                   offset: statuses.count))
+      if let account {
```
Dimillian, line 157: I'm ok with the logic, should be refactored into a proper function like processStatuses or something, later we'll probably want to add more local stuff.

Packages/Models/Sources/Models/Status.swift
```diff
@@ -59,6 +59,8 @@ public struct Status: AnyStatus, Decodable, Identifiable, Equatable, Hashable {
     id + createdAt + (editedAt ?? "")
   }
 
+  public var shouldHighlight: Bool?
```
Dimillian, line 62: To separate that a bit, I would make a new protocol like protocol `StatusUI` and add this the props inside. Then make `Status` conform to this new protocol. Name the property something like `ui_shouldHighlight`. I know it's not pretty but that way it should be crystal clear what is backend / models status and UI extensions for the app.

Packages/Timeline/Sources/Timeline/TimelineView.swift
```diff
@@ -75,6 +75,7 @@ public struct TimelineView: View {
       if viewModel.client == nil {
         viewModel.client = client
         viewModel.timeline = timeline
+        viewModel.account = account
```
Dimillian, line 78: You don't need to pass the Current account env here. You can directly use `CurrentAccount.shared`, it's fine as a singleton.

Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
EvilOne: Totally agree! On my list before remove the draft status.

Packages/Models/Sources/Models/Status.swift
EvilOne: I like this better than what I have. Will try to come up with a good variable name.

Packages/Timeline/Sources/Timeline/TimelineView.swift
EvilOne: Missed the singleton. Thanks!

PR #595, opened 2023-02-01
Timeline highlighting

EvilOne:
I let my other branch fall too far behind, and it was easier to create a new PR.

### DIFF
--- a/Packages/Models/Sources/Models/Status.swift
+++ b/Packages/Models/Sources/Models/Status.swift
@@ -59,6 +59,8 @@ public struct Status: AnyStatus, Decodable, Identifiable, Equatable, Hashable {
     id + createdAt + (editedAt ?? "")
   }
 
+  public var shouldHighlight: Bool?
+
   public static func == (lhs: Status, rhs: Status) -> Bool {
     lhs.id == rhs.id
   }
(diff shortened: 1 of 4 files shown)


Packages/Models/Sources/Models/Status.swift
```diff
@@ -54,11 +54,17 @@ public protocol AnyStatus {
   var language: String? { get }
 }
 
-public struct Status: AnyStatus, Decodable, Identifiable, Equatable, Hashable {
+protocol StatusUI {
+  var uiShouldHighlight: Bool? { get set }
```
EvilOne, line 58: I couldn't think of a better variable name! 😿

Packages/Timeline/Sources/Timeline/TimelineViewModel.swift
```diff
@@ -71,7 +76,13 @@ class TimelineViewModel: ObservableObject {
        !statuses.contains(where: { $0.id == event.status.id })
     {
       pendingStatusesObserver.pendingStatuses.insert(event.status.id, at: 0)
-      statuses.insert(event.status, at: 0)
+      var newStatus = event.status
+      if let accountId {
```
EvilOne, line 80: This is a one-off case where it is just a single event.

Dimillian:
Nice!

Issue #398, opened 2023-01-26
Bug: Un-gender the German translation


dertuxmalwieder:
Not quite a bug, but room for discussion, I guess:

The German language does not have a "gender-neutral" word for "users", we usually use generic masculine nouns ("users" = "Nutzer", which is homonymic with "male users"). In recent years, a few attempts have been made to establish a neutral form, e.g. "Nutzer*innen" or "Nutzer_innen" or "Nutzer:innen", which adds the female ending "-innen".

Now the problem is that these disturb screen readers - using these forms would sound *weird*, to say the least, for severely visually impaired people. When doing https://github.com/Dimillian/IceCubesApp/pull/397, I saw that @vollkorntomate used one of the several "neutral" forms.

Personally, I would prefer to have "Nutzer:innen" corrected to "Nutzer", but I don't want to start an *edit war* on that. Just a suggestion ... :-)

vollkorntomate:
As far as I know is VoiceOver perfectly capable of reading gender-neutral words like this. iOS even uses the same style with the colon (`:`), e.g. in the system settings.

But anyway, I don't have a strong opinion in either direction.

jue89:
iOS VoiceOver pronounces the so called Glottisschlag, which is introduce by the ':', perfectly fine. From the POV, there's nothing to be fixed.

I'd recommend to keep it like it is - it explicitly includes non-maskulin people, whereas the "Generisches Maskulinum" does this only implicitly.

It may feels cumbersome to some people, as it arose during the recent years and may feel unfamiliar, but I'd say it's worth it. Language is always evolving. And if it includes people actively, it's a good development. Language is a tool and not art, that must be preserved.

sothawo:
> we usually use generic masculine nouns

that's exactly what is the whole point of the gender discussion. Assuming that the masculine form is enough for all. There are many people who don't feel themselves addressed by this. So I would stick with the gendered variant.

dertuxmalwieder:
Is anyone here who would *not* feel addressed?

sothawo:
The question is not about people _here_. Developers are - alas - mostly male. It's about the users of the app. But I'll not discuss this further, sad it still comes up in 2023.

dertuxmalwieder:
You claim that "there are many people who don't feel themselves addressed by this“, and I wonder how many of those use Ice Cubes. I mean, they'd probably show up here, right?

Suggestion: Let’s ask the affected users, not discuss this among us.

[...]

PR #597, opened 2023-02-02
Updated German localization

dertuxmalwieder:
- translated a few more strings
- unified "participants"


jue89:
... and he silently un-gendered "Teilnehmende" ... /cc #398

Dimillian:
That the kind of stuff I can't check. I guess we should define owner for localization soon so I'll not merged before an LGTM from the owner.

dertuxmalwieder:
It would have to be “Teilnehmer:innen” to be consistent, so it was wrong either way…

jue89:
I recapped my PoV in the related issue. I just wanted to create transparency which hasn't been done by @dertuxmalwieder ("unified" instead of "changed to generic masculine").

I won't continue these edit wars - it's eating up valuable time. And personally I'm not affected. But I have friends who would ... that's why I think gender-neutral language is worth to be considered.

Issue #576, opened 2023-01-31
Bug: Incorrect rendering of post


pmonks:
[This post](https://cinematheque.social/@12pt9/109786219724322016) is not being rendered correctly, as shown here:

![image](https://user-images.githubusercontent.com/54865/215903846-dd773481-e5b8-4df3-8c32-ab55e3bb2687.jpeg)

Note: other clients I checked render it correctly.

Dimillian:
@xurble Do you have an idea? It's probably not your change but just the current lib failing at converting the HTML. I sax that from time to time.

xurble:
Just looking at the post, the first thing that jumps out is that the html is malformed. It truncates in the middle of an open anchor tag. 

Could that be it?

xurble:
Looking at the thread, there's mention of zero width joining characters. I'll try and take a look this evening. 

I'd actually like to take another crack at the markdown generation.

EvilOne:
This has been happening before any of @xurble changes. Is there an issue rendering due to EmojiText?

EvilOne:
Here is one that I bookmarked a long time ago that has an issue displaying -> https://newsie.social/@jeff/109716493698039652

It renders like this:

<img width="50%" src="https://user-images.githubusercontent.com/169561/216196327-9f1fc267-34d3-4562-86d6-0c3b6ff66ecf.png">

xurble:
FWIW the zero width joiners were breaking HTML2Markdown

EvilOne:
Nice find!

sothawo:
Are there different kinds of zero width joiners? This screenshot show that my post  is rendered correctly, there is a zwsp between "#tag" and "s" 

![image](https://user-images.githubusercontent.com/8456476/216320499-bff9db0e-5140-4f40-b452-f237cdf49618.png)

sothawo:
Addition: I enter it as a unicode character, not as an html entity

xurble:
Can you post the url for the screenshotted post please?

sothawo:
https://digitalcourage.social/@sothawo/109795642095913757 new one, the original was not (only visible for me)

sothawo:
well, it didn't break the current version - and did not break previous ones, I use always use this when I have pluralized tags.

xurble:
@sothawo 
Your working zero width is (hex) `80 8B`
The breaking one is `E2 80 8D`

I don't know why HTML2Markdown chokes on one but not the other.

[...]

PR #610, opened 2023-02-02
Newer, better, faster HTML to Markdown converter

xurble:
It removes all the hacks and just uses the SwiftSoup parser

It will fix issue #576 and also an unlogged issue with the old implementation that could corrupt urls with underscores in them.


xurble:
As a side effect, you can remove the HTML2Markdown dependency, which I haven't done in case its needed elsewhere.

Dimillian:
This look epic but I'm afraid we would need to test it on a lot of statuses?

xurble:
God yes!
It's on my phone at the moment.  It seems to fix all the known problem cases I have bookmarked.

EvilOne:
@Dimillian @xurble to really test this change out, assuming what I propose is simple, how about an experimentation area in settings? Within there we would have render with new algorithm (not really suggestion this exact text) which would act asa feature flag. This feature flag, when active, would run this code instead of the existing code. The default would be off. 

Then when/if released like this can call it out in the release notes and such. 

Thoughts?

Dimillian:
@EvilOne I'm waiting to fix the crash on timeline refresh so I can start the 1.4.0 train, my plain is to merge this as soon as 1.4.0 hit and use it for a few days to see if there is any regression. From what I've tested so far, I couldn't not find. 
I agree that a feature flag would be ideal but it complicates a bit the process

PR #622, opened 2023-02-02
Updated German translation

dertuxmalwieder:
Added one translation, re-gendered participants (in response to #398 - I was wrong).


jue89:
It may looks like I have no hobbies :D ... but you forgot this line:

https://github.com/Dimillian/IceCubesApp/blob/5f4d7b5eafe94078ee4adb487e08f92785a10e6a/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings#L293

Issue #599, opened 2023-02-02
Feature request: Add Open Dyslexic font


markrprior:
It might be appreciated if you add the Open Dyslexic font to the fonts available, either by simply adding it to the list or by making it a specific option so that it is more obvious. Toot! for example has it as a specific option but noting its availability may be enough to improve accessibility. See [https://opendyslexic.org/](https://opendyslexic.org/)

Fastidious:
Can’t it be installed by the user as a custom font?

markrprior:
> Can’t it be installed by the user as a custom font?

I can't see a way of doing that in the iOS app. The custom "Timeline Font" option is a list of fonts to select from rather than a way to add one. In any case you want to make it easy for a non technical user to be able to use it rather than make them jump through hoops.

PR #626, opened 2023-02-03
Fonts

xurble:
Addresses #599 

This installs 2 custom fonts: OpenDyslexic and also Atkinson Hyperlegible for accessibility reasons.

But to satisfy the licensing requirements, it also adds a basic About screen to settings with a link to those fonts and, for completeness, all the open source libraries Ice Cubes uses.


Dimillian:
Unexpected About screen but super nice!

Dimillian:
So, on device, the font picker controller don't display those two fonts :/ @xurble can't really understand why.

xurble:
Sorry about that.  Obviously it worked in testing.  If I had to guess, I think it will be this setting.

Screenshot 2023-02-04 at 08 09 28

xurble:
I would check "Install Fonts"

Dimillian:
Let's try.

PR #627, opened 2023-02-03
Japanese translation optimization

nekolaboratory:
In terms of tab UI, this is cleaner.

### DIFF
--- a/IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
@@ -137,7 +137,7 @@
 "settings.other.autoplay-video" = "動画を自動再生";
 
 // MARK: Tabs
-"tab.explore" = "検索";
+"tab.explore" = "探索";
 "tab.federated" = "連合";
 "tab.local" = "ローカル";
 "tab.messages" = "メッセージ";
--- a/IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
@@ -137,7 +137,7 @@
 "settings.other.autoplay-video" = "動画を自動再生";
 
 // MARK: Tabs
-"tab.explore" = "エクスプローラー";
+"tab.explore" = "検索";
 "tab.federated" = "連合";
 "tab.local" = "ローカル";
 "tab.messages" = "メッセージ";


Dimillian:
@kum4423 should this be merged?

nekolaboratory:
A note for translators:
The character strings in the lower tabs take up too much space in full-width characters in Japanese.
Also, since it is a search screen rather than an explorer, it is translated as "検索".
In fact, many clients are labeled "検索".

kum4423:
Frankly, I disagree.
Mastodon, unlike Twitter, uses "explorer" instead of "search" because of its federation feature.
It's true that many client apps search and translate people, but that's probably because they don't know how to translate properly. I think "explorer" is a good way to describe it.

nekolaboratory:
I agree with some of your opinions.
However, the only problem this time is that if the character string of "エクスプローラー" in the lower row is Japanese double-byte characters, the width is too wide and it stands out unnecessarily.

So, I didn't fix the translation on the destination explorer screen.

Also, it is a little unnatural that the icon is a search even though it is called Explorer.

nekolaboratory:
In addition, Japanese people are accustomed to the dialect of "検索" used by many clients.
I think there is an advantage to valuing such intuition.

kum4423:
As a Japanese, I understand what you are saying. However, this application is originally a client application of Mastodon, and its main source is "エクスプローラー". I think it is wrong to translate it into "検索" just because it is familiar and easy to understand for Japanese people. I think "search" is correct for the UI, but from the perspective of UX and translation, I don't agree with it. Please could you make a concession here?

kum4423:
> @kum4423 should this be merged?

Personally No. It has to be combined with other areas.
However, if someone else insists that this must be a "search", I'll leave it to you, Thomas...

nekolaboratory:
I think your translation is correct too.
However, what is worrisome is that "Explorer" is too long as a double-byte character string, so when checking in the iPhone 13 Pro environment, the label is resized and rattled.

Why don't you use the Japanese translation of "探索" if you use the correct expression, "エクスプローラー"?
It solves the UI problem.

If you agree with this proposal, I will re-modify this change as "探索".

tab

kum4423:
Do you know the difference between a literal translation and an implied translation?
A literal translation would be "search," but that is not the case here.
If you are concerned about the length of the notation, I think "見つける" is still more appropriate than "探索".
How about this?

kum4423:
I should also mention that the other language settings still conform to the "Explorer" word set.

nekolaboratory:
I think I understand that. So I'm not trying to force a literal translation.
"見つける" looks a little childish.
I think one of "検索", "探索", or "捜索" seems appropriate for this tab arrangement.

kum4423:
How is "見つける" childish?
Have you ever been born and raised abroad? I don't think so.

Unless the "見つける"or "探索" proposal is approved, there is no more room for discussion, so I will leave the rest to the developer's decision. I don't like to argue.

Which one will you adopt?

nekolaboratory:
In Chinese, the word “探索” is used.

```
"tab.explore" = "探索";
```

kum4423:
> @kum4423 should this be merged?

You can take a look at our conversation so far and decide if you want to change the word "explorer". I don't want to discuss this further on a Saturday holiday.

nekolaboratory:
Among them, the word "探索" fits nicely in the row of tabs.

Dimillian:
Explore is more in line with the rest of the application and Mastodon. Thanks for the discussion.

PR #629, opened 2023-02-03
Fix translate button for boosts in context menu

vollkorntomate:
Boosted posts would not show the correct "Translate" button in the context menu. Instead, the booster's default language decided if the button was visible.
This PR fixes this.

Screenshot of how it was before (expected button text: "Translate from Spanish"):
Simulator Screen Shot - iPhone 14 Pro - 2023-02-03 at 11 38 09


Dimillian:
Nice catch!

PR #633, opened 2023-02-03
Trying to recognize the language of the Toot to set it correctly if none is provided

Dean151:
Probability threshold arbitrary set to 85%.

Seems to work enough for French, English and Spanish. Also tested and with only a few words, probability is low unless there is a very language specific word

It's using Natural Language API from apple, so it works offline, and with zero cost

### DIFF
--- a/Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
+++ b/Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
@@ -1,6 +1,7 @@
 import DesignSystem
 import Env
 import Models
+import NaturalLanguage
 import Network
 import PhotosUI
 import SwiftUI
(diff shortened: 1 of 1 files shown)


Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
```diff
@@ -126,6 +127,21 @@ public class StatusEditorViewModel: ObservableObject {
...
+
+      if !hasExplicitlySelectedLanguage {
+          // Attempt language resolution using Natural Language
+          let recognizer = NLLanguageRecognizer()
+          recognizer.processString(statusText.string)
+          // Use languageHypotheses to get the probability with it
+          let hypotheses = recognizer.languageHypotheses(withMaximum: 1)
+          // Assert that 85% probability is enough :)
+          // A one word toot that is en/fr compatible is only ~50% confident, for instance
+          if let (language, probability) = hypotheses.first, probability > 0.85 {
```
Dimillian, line 139: My only concern is that it can overwrite the user default setting. And also at the time of posting, if the language detected is different, then the UI is wrong.

Dean151: I can try to make the switch dynamically when the recognizer have enough probability? And the switch would be visible in the bar for the user. And with a little "pop" animation, the user see it happening?

I can manage to make the changes tonight

Dimillian: But it means it would have to be run every time the user type. I'm more afraid of the performances issues then. I think it could be fine as is, let's merge and see the feedbacks.

Dean151: And/Or a switch within the settings ;)

PR #645, opened 2023-02-04
Italian Localization Update: Settings & Info

torre76:
- In App Browser Reader View
- Open Source Software used


Dimillian:
Now you need the new haptics settings :)

PR #648, opened 2023-02-04
Update Localizable.strings

enetter:
Added missing translations in French, mainly in the haptic section. I've tried to stock to the vocabulary already used by other translators.


Dimillian:
Merci !

enetter:
Avec plaisir ! Bravo pour l'application que j'utilise tous les jours.

PR #661, opened 2023-02-05

Jerry23011:
### Changes
- Add translation for browser extension

@nixzhu could you check it out :)


nixzhu:
LGTM

PR #662, opened 2023-02-05
Add muting duration submenu

### LINKED ISSUE #201: Feature request: Support variable length mute periods
From the Mastodon web interface, you can mute a user for a variable length of time. This would be a very useful feature to implement in the app.

sothawo:
This PR adds a submenu to the menu when muting an account to specify the duration of the muting. The values are the same that are offered on the Mastodon website for this operation. The selected value is added to the mute API call.

@Dimillian I have this as draft PR, because I have 2 questions:

1. Is it ok where I put the enum for the different values or do you have a special place where you collect such enums?
2. While implementing this I saw that there is a menu entry "Enable Notifications/Disable Notifications@ directly below the mute entry. After checking the code and the API I saw that this is not the same as the button in the account detail view with the bell that enables or disables notifications. This API rather is meant to enable notifications from a user I am following when I am muting her (don't show the statuses but notify). It is not possible to set this anywhere in the Web UI and as this uses the mute API, we would need the duration selection there as well. Probably might need to reword the entry to "mute but keep notifications".  As I find this entry a little confusing (why mute but keep notifications) and as it is not available in the Web UI, I commented this out to remove that entry from the menu. Would that be ok, or do you have a better idea to name this?

image

Closes #201


Dimillian:
1. Right, this enum sound good in its own file next to where it's used. Better than inline the view. 
2. I think there is some mixup in term of menu + actions in there, let me check.

Dimillian:
All good: https://github.com/Dimillian/IceCubesApp/commit/b598434648be91548fc91fd2570fef75c7bece95 the menu action were definitely wrong

sothawo:
I moved the enum to it's own file and reverted my changes on the notify menu entry to the last changes from main

IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
```diff
@@ -24,6 +24,14 @@
 "enum.avatar-position.top" = "トップ";
 "enum.avatar-shape.circle" = "サークル";
 "enum.avatar-shape.rounded" = "ラウンド";
+"enum.durations.infinite" = "infinite";
```
nekolaboratory, line 27: These are expressed in Japanese as follows.

"enum.durations.infinite" = "無期限";
"enum.durations.fiveMinutes" = "5分間";
"enum.durations.thirtyMinutes" = "30分間";
"enum.durations.oneHour" = "1時間";
"enum.durations.sixHours" = "6時間";
"enum.durations.oneDay" = "1日間";
"enum.durations.threeDays" = "3日間";
"enum.durations.sevenDays" = "7日間";

sothawo: thanks, added that

Dimillian:
It's all good to me, will merge this evening for 1.4.1!

Issue #654, opened 2023-02-05
Feature request: Remove tab string from iOS app


nekolaboratory:
discussion
Is the tab title displayed only in the iOS app really what you want?
This is a Japanese environment, but the character string in the explorer is too long and the UI collapses.
In the past, I proposed shortening the string to solve this problem(#627), but I came to the conclusion that it is unnecessary if it is only displayed in the iOS app in the first place.
We keep in mind that this can occur in multiple languages ​​even if only the Japanese environment is resolved.

PR #669, opened 2023-02-06
Remove tab string from iOS app

nekolaboratory:
Fixed an issue where the display of tab menu names may be corrupted on iOS devices with small screens.

Closes:#654


sothawo:
Instead of always removing the labels, I'd think to have this configurable (show labels with icons) would be a better way. Especially for new users who open the app for the first time, it might not be clear what these icons are without the explanation by the labels.

Dimillian:
Definitely need to be a setting. And I've yet to pull it to see how it look but I hope iOS adapt the image bottom margin then.

nekolaboratory:
I was of that opinion for a while.
However, depending on the language and label size, the characters may be cut off and it does not make sense.

I also refer to the fact that the Mastodon official app does not have an icon label by default.

Tab

sothawo:
This is a very good example that making this configurable is the best way. For a user where the label is long and who has a small display it can be disabled. But I'd keep the label as default for everybody else.

nekolaboratory:
Here is a screenshot of the simulator.

Simulator

Explicitly showing functions to users with strings is important, but considering the accessibility of people who use languages ​​that cannot read the default English and have not been translated, UI that can be conveyed only by icons is important.

echidistanti:
Having it as an option is the optimal solution. Usability guidelines recommend labels, but we're talking about icons that, in this case, have universally recognized functionality, true pictographs

Dimillian:
I like the idea, again. But this create a huge UI problem. The icons should be center aligned, and we can't do that in SwiftUI until we make a custom tabbar. 
I would be ok to merge it if it's an option in display settings, with a default to true (showing the label). Until then I can't merge it.

nekolaboratory:
I understand your opinion.

Then we plan to find time and make changes to add a show/hide option to the settings screen.

nekolaboratory:
@Dimillian
Discussed tab label display settings

What do normal users call the string under the tab icon?

Currently, it is set to Tab Name.

Dimillian:
All good thanks!

PR #675, opened 2023-02-06
Update SideBarView.swift

jiannazzone:
Change .edgesIgnoringSafeArea(.top) -> .ignoresSafeArea(edges: .bottom) on line 135 (Divider). On iPad with SideBarView, the divider would cut into the status bar clock


Dimillian:
Not really pleasing aesthetically, even on iPad. And then it also show on the space in macOS. I guess something more custom would have to be made.

PR #678, opened 2023-02-06
Clean up HapticManager

EvilOne:
1. The view now only knows that it wants to _fire_ a haptic event.
2. The HapticManager verifies the user's settings before firing a haptic event.
3. Haptic setting menu will **not** show up if the device does not support haptics (for instance, an iPad).

### DIFF
--- a/Packages/Env/Sources/Env/HapticManager.swift
+++ b/Packages/Env/Sources/Env/HapticManager.swift
@@ -27,4 +27,8 @@ public class HapticManager {
   public func notification(type: UINotificationFeedbackGenerator.FeedbackType) {
     notificationGenerator.notificationOccurred(type)
   }
+
+    public var supportsHaptics: Bool {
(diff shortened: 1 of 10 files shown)


Dimillian:
This is amazing, I did that part very quickly on holiday and it was definitely trash tier. Thanks for making it so much better!

Packages/Env/Sources/Env/HapticManager.swift
```diff
@@ -1,30 +1,73 @@
...
+  @MainActor
+  public func fireHaptic(of type: HapticType) {
+    guard supportsHaptics else { return }
+
+    switch type {
+    case .buttonPress:
+      if userPreferences.hapticButtonPressEnabled {
+        impact()
+      }
+    case let .dataRefersh(intensity):
```
Dimillian, line 35: dataRefersh -> dataRefresh

seanlifetime: 😄 oops

PR #687, opened 2023-02-06
Fix custom emoji with underscores

### LINKED ISSUE #611: Bug: Custom emojis not shown
My instance has a few custom emojis, but when clicking the custom emojis button in a new post, I get no emojis at all, let alone my custom emojis.

Bonus (really an enhancement request I think?): that button should really show all regular emojis as well.

xurble:
I've given this a run around the block and it seems OK.

There is probably a single regex that can do this whole thing, I can't write it!

Fixes #611


Dimillian:
At this point, you're a Regex wizard I gave up on them long time ago 😆

Issue #581, opened 2023-02-01
Feature request: Timeline: restore the selected timeline at the next startup


nekolaboratory:
It feels a little inconvenient that the selection of tabs on the main timeline is initialized at startup.
I want to keep the selected kind of timeline.

jsynowiec:
Perhaps related to #561?

alper:
I don't see how this is done. The thing that's preventing me from using Icecubes as a daily driver is to have Home open on the same list that I left it if I start the app.

alper:
Is there an open ticket tracking having a List as your starting timeline?

nekolaboratory:
I have never seen it, I guess.

PR #694, opened 2023-02-07
Timeline: restore the selected timeline at the next startup

nekolaboratory:
Closes: #581


Dimillian:
I think you need to rebase on main one more time, missing the `remoteLocal` new filter parameter.
It create when switching account for example, if you saved a list not available on the other account etc... 
And it'll also fail whenever I'll add new filters (but I guess this is fine). 
Also have to test the various login / logout configuration. 

I just think it's a lot of effort for such a small feature, and it could create unforeseen feature. Multi accounts being a problem right now with this feature.

nekolaboratory:
I will fix that. Thank you!

nekolaboratory:
Please review again.

nekolaboratory:
https://github.com/Dimillian/IceCubesApp/pull/694

nekolaboratory:
Please review again.

nekolaboratory:
@Dimillian Regarding the account switch problem that you are concerned about, it is a little overbearing, but how about saving the state by limiting it to the tabs common to the account as follows?

```
    .onChange(of: timeline) { timeline in
      if(timeline == .home || timeline == .federated || timeline == .local) {
        lastTimelineFilter = timeline
      }
    }
```

Dimillian:
@nekolaboratory Sound good. Unless it really break the feature as you would use it.

nekolaboratory:
Thank you. Changes have already been committed.

Dimillian:
Thanks! Let's give it a try!

nekolaboratory:
Cheers!
On the Mastodon instance I'm based on, the local timeline is mainstream.
So being able to record the default timeline is a very useful feature.

Can't wait for the release!

PR #709, opened 2023-02-07
Fix Conversation.lastStatus nullability issue

### LINKED ISSUE #696: Bug: Cannot load any private message
I can't show any private messages, I always get a fullscreen error message.
Tapping retry doesn't help.

image

I have some non linear private messages. So not every message was a reply to the oldest message. Some private mentions have multiple replies - like normal messages. Maybe this is the issue.
Showing those messages in Mammoth results in incomplete threads (some are filtered out)

sothawo:
#696 comes up when the /conversations endpoint returns a conversation where the `last_status` property is nil. This is allowed according to the [API documentation](https://docs.joinmastodon.org/entities/Conversation/#last_status).

I changed the corresponding property in `Conversation` to `public let lastStatus: Status?`, the rest of the code changes are adaptions to this now optional property.

Closes #696


Dimillian:
Thanks, just added back some UI to where it was :p

PR #719, opened 2023-02-08
Updated German translation

dertuxmalwieder:
more strings

### DIFF
--- a/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
@@ -82,7 +82,7 @@
 "settings.display.theme.systemColor" = "Systemeinstellung verwenden";
 "settings.general.browser" = "Browser";
 "settings.general.browser.in-app" = "In-App-Browser";
-"settings.general.browser.in-app.readerview" = "In-App Browser Reader View";
+"settings.general.browser.in-app.readerview" = "In-App-Browser-Lesemodus";
 "settings.general.browser.system" = "Systembrowser";
 "settings.general.display" = "Anzeigeeinstellungen";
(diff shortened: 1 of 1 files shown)


IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -82,7 +82,7 @@
 "settings.display.theme.systemColor" = "Systemeinstellung verwenden";
 "settings.general.browser" = "Browser";
 "settings.general.browser.in-app" = "In-App-Browser";
-"settings.general.browser.in-app.readerview" = "In-App Browser Reader View";
+"settings.general.browser.in-app.readerview" = "In-App-Browser-Lesemodus";
```
icod, line 85: I believe that in the German localization of Apple software the Safari Reader is called "Reader" in German too. 

https://applelocalization.com/?q=reader&l=English&l=German

dertuxmalwieder: It is, but other browsers use “Lesemodus” or similar terms. Personally, I’d always prefer actual translations instead of just declaring English words “fixed terms”..?

icod: Me too! Apparently, Apple regard the term “Safari Reader” so quintessential and untranslatable (for marketing reasons?). 

My opinion? Is Safari the in-app browser, then it should be "Reader" for consistency-sake. If the in-app browser is a white-label browser offering a reader mode, then Lesemodus is the better choice.

I'll leave the final decision with you, of course. This is just my two cents 🙂

dertuxmalwieder: The in-app browser is Ice Cubes. ;-)

Issue #715, opened 2023-02-08
Feature request: mentions highlight color


theotherlinh:
It be nice to have a dedication mentions color that was separate from the secondary background color. Only to better highlight them in my feed. It will stand out better from preview cards using the secondary background color.

Issue #553, opened 2023-01-31
Feature request: Differentiate direct messages more strongly in the Home feed


pwinn:
You're already showing a different icon for direct messages, so you know they're direct messages while building that list. It's still deeply disconcerting to see them appear in the Home feed just like public messages, and have only that tiny icon to let me know I didn't accidentally post semi-private things publicly. (Yes, I know DMs aren't *truly* private, at least from instance owners.)

Could those posts have a colored background and be indented, like they are in the Direct Messages conversation view? At least *something* to make them stand out and be immediately obvious they're DMs. 

Thanks.

theotherlinh:
Agreed. This should also be different from general mention highlights as well. Ideally just another optional color.

PR #720, opened 2023-02-08
Add direct status highlighting

EvilOne:
This opens the door for future enhancements to add a direct message and/or mention colour to the existing themes system. 

I am not good at picking individual colour choices, so I didn't add additional items to the theming system.

*This PR sets things up to implement #553 and #715.*

CleanShot 2023-02-08 at 05 36 14@2x
CleanShot 2023-02-08 at 05 35 24@2x

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
@@ -3,6 +3,8 @@ import Models
 import Network
 import SwiftUI
 
+import DesignSystem
+
 @MainActor
 public class StatusRowViewModel: ObservableObject {
   let status: Status
@@ -30,6 +32,8 @@ public class StatusRowViewModel: ObservableObject {
 
   @Published var favoriters: [Account] = []
   @Published var rebloggers: [Account] = []
+
+  private let theme = Theme.shared
   
   var seen = false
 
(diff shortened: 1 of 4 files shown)


Dimillian:
I think picking the exact same color than the direct message one in conversation view (I think it's tint color with some opacity is fine). It's not something we should offer settings on, too easy to mess it up.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -30,15 +32,24 @@ public class StatusRowViewModel: ObservableObject {
...
   var filter: Filtered? {
     status.reblog?.filtered?.first ?? status.filtered?.first
   }
 
-  var shouldHighlightRow: Bool {
-    status.uiShouldHighlight != nil
+  var highlightRowColor: Color {
+
+    if status.visibility == .direct {
+      return theme.tintColor.opacity(0.15)
```
Dimillian, line 47: Looking good

EvilOne: I think it is subtle enough and indicates that this is a direct message instead of a highlight.

phpmacher:
oh, i hoped for a setting

Dimillian:
We add a settings and then the next 10 bugs will be "I can't see the DM in my Timeline" because the user will have set the background color to something unreadable. Will see when I'll do the font color settings.

phpmacher:
@Dimillian 

1. you can't make anything idiot-proof because idiots are so creative

2. you could theoretically check the chosen colors for similarity and display a hint if necessary

3. doesn't that apply to all settings?

Dimillian:
It does 😆 Anyway, I think more theming options is still planned, so I guess customization of those colors will go in this phase. Don't know when it'll be yet.

PR #723, opened 2023-02-08
Fix handling of polls

### LINKED ISSUE #332: Bug: White Text for Polls in Light Modes is Unreadable
E369AEEE-CE40-4927-BA72-BD3AB1F8FE5D
97E140E4-E292-46BF-A549-330E27656673

### LINKED ISSUE #625: Feature request: multiple choice polls
Currently, multiple choice polls are treated as single choice ones. There is no way to select multiple options.

sothawo:
This PR fixes the incomplete handling of polls and enables working with multi-choice polls.

I made the following changes:

For both kind of polls, all options are displayed in rows as before, multiple choice and single choice options look like the classical checkboxes or radio-buttons. The user can select different options, the behaviour is like usual for these types of buttons.

As long as it is possible to vote, there is a `Send Vote` button below them, which will post the poll to the server.

image

When it is not possible to vote, either because the user has already voted or the vote is expired, the send button is removed.

image

Closes #625 
Closes #332

### DIFF
--- a/IceCubesApp/Resources/Localization/ca.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/ca.lproj/Localizable.strings
@@ -384,6 +384,7 @@
 "status.poll.duration" = "Durada de l'enquesta";
 "status.poll.frequency" = "Freqüència de l'enquesta";
 "status.poll.option-n %lld" = "Opció %lld";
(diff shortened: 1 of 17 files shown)


Dimillian:
Very nice thanks!

PR #724, opened 2023-02-08
Add translate button to the preview post in settings

xurble:
So this is why I did the preview toot in Latin.

Because, the "Show Translate Button" is a display setting, it makes sense to show it, but then you have the hassle of is it currently in English or has it been localised, so ... Latin.


seanlifetime:
But showing the translate button is optional, so the preview happens to have the option off. 😉

Dimillian:
Thanks for resolving the conflicts!

PR #737, opened 2023-02-09
Attachments on preview

xurble:
I'm not sure about this.  There is a setting to preview but it's kind of busy looking and takes up a lot of space.


xurble:
Just realised I don't know how to request a review.

xurble:
I ballsed this up.  It's branched from the translate change so you can't apply it without applying the other

PR #738, opened 2023-02-09
Attachments on preview

xurble:
I'm not sure about this one. 

One the one hand, there's a setting you can now preview.  On the other hand it takes up a lot of space and it's a bit busy.


Dimillian:
I've tried a bit but yes, it's a bit to tall and too busy to really be useful I think :/

xurble:
That's what I thought. People know what large and compact means I think.

PR #739, opened 2023-02-09
Fix iPad display settings background color

J-rg:


Dimillian:
Thanks but this is already fixed on main.

PR #748, opened 2023-02-09
Update Localizable.strings

kum4423:
Restore the original notation as the uniformity of translation will be lost.


kum4423:
Even if I fixed it, I felt that there was no point in doing so, since it is certain that someone in particular will localize it differently again for some reason or another.

Issue #756, opened 2023-02-09
Feature request: Display username in DM


nekolaboratory:
In my community, there are people who either have no screen name or change it frequently.

By referring to the username, it is possible to track in the timeline, but not in DM.

Please display username in DM as a remedy.

Dimillian:
See https://github.com/Dimillian/IceCubesApp/commit/dedc7af49ccbe2d73f248a77be163ac66e193daa

PR #758, opened 2023-02-09
Display username in row of Conversations list

nekolaboratory:
Before:

Simulator Screen Shot - iPhone Pro - 2023-02-10 at 07 10 00

After:

Simulator Screen Shot - iPhone Pro - 2023-02-10 at 07 20 00

Closes: #756


Dimillian:
I've fixed it by returning the username if it's empty now, should do the trick for your case. This UI is too problematic for group DM.

Issue #304, opened 2023-01-23
Bug: GIF-based avatars don't display


yankcrime:
My avatar was uploaded to my Mastodon instance as a GIF, and whilst this displays fine in the site and other clients in Ice Cubes it's just blank.  If I change to a PNG or a JPG it displays just fine.

sothawo:
another example: @FARBfernsehen@mastodon.social, doesn't render in icecubes

Havhingstor:
That's a problem with the Image Engine Nuke, which is used for the profile images (https://github.com/kean/Nuke). That bug is now fixed, but that fix isn't yet implemented in Ice Cubes, as there is no new version of nuke. Additionally, the version of Nuke used in Ice Cubes isn't the current one, so that has to be updated too.
@Dimillian You could obviously directly reference the commit (https://github.com/kean/Nuke/commit/df63473487c6b716247f0c42630ba4b6aa7b68b1) or main-branch of nuke, but I guess you'd probably rather wait until @kean bumps the version number.

Havhingstor:
See https://github.com/kean/Nuke/issues/668

Dimillian:
Awesome!

kean:
I push version [11.6.2](https://github.com/kean/Nuke/releases/tag/11.6.2). Thanks, @Havhingstor, for finding the root causing and fixing it. It's also worth fixing it on the [Gifu](https://github.com/kaishin/Gifu) level for folks who are using it directly. I'll try to take care of it when I can unless someone else does it earlier.

PR #759, opened 2023-02-09
Update to the newest version of nuke

Havhingstor:
Static gifs as avatars are now rendered correctly. (See also #304)
Examples: https://social.bund.de/@bfdi, https://mastodon.social/@FARBfernsehen


Dimillian:
Yeah!

PR #776, opened 2023-02-10
Add accessibility labels to buttons in compose screen.

TheQuinbox:
I've added .accessibilityLabel modifiers to all the buttons in the compose 
screen that needed it, other than the OpenAI features one, as I couldn't 
find the code for it, and am not entirely sure what it does.

Do I have to add these strings to a localization file somewhere? 
apologies, I'm very new to Xcode and iOS development still.


Dimillian:
Awesome! Thanks! 

Yes it would be nice if you could create a key for each, look at Localizable.strings and add them here. You can create a new prefix for accessibility keys. Like accessibility.editor.button.drafts = "Drafts"; for exemple. You need to add them in all files. Leave it in English for now everywhere and then it'll be translated later.

TheQuinbox:
Alright, all done :)

IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -421,3 +421,10 @@
 "filter.contexts.profiles" = "Profiles";
 "filter.action.warning" = "Hide with a warning";
 "filter.action.hide" = "Hide completely";
+
+"accessibility.editor.button.attach-photo" = "Attach photo";
```
Dimillian, line 425: You need to copy that in all localizable.strings files :D

TheQuinbox:
Oh, okay i see. Sorry, new like I said haha.
Also gone ahead and created a marker :)

PR #785, opened 2023-02-11
swipe actions improvements

sothawo:
ok, changes for swipe actions:

## settings

I added a new settings for Swipe actions and in there a section for swipe actions on posts (I can think of more places where it might be handy to configure swipes)

Screenshot 2023-02-11 at 13 41 31
Screenshot 2023-02-11 at 13 43 10

I'm not quite content with the wording for the 4 swipe entries (_swipe left outer_, _swipe left inner_) but have no better idea currently. It would be nice as well to have the icons in the picker entries, probably need to add them with the Text in some view.

I have defined an enum for these values that can provide the symbol names and display label keys, perhaps it might make sense to use this at other places as well so we would have the labels and icon names in one place to maintain; but that should be a different PR then.

I store this 4 values in the user preferences, would it be better to put them in a struct and serialize/deserialize that when the preferences are written?

## display

For the outer one I used the tintColor, the inner one is gray. Alas it seems not to be possible to have the icons in a not-filled style, I would have liked to show the state of bookmarked faved etc in there. And I did not find a way to change the color of the icon/text in the swipes to something different than white.

The text is in the swipes for me during development, should I take it out or can we leave it in there?

left side

Screenshot 2023-02-11 at 13 44 09

right side

Screenshot 2023-02-11 at 13 44 19

Localizations for all languages besides en have not yet been added, waiting until all necessary changes are done

### DIFF
--- a/Packages/Env/Sources/Env/StatusAction.swift
+++ b/Packages/Env/Sources/Env/StatusAction.swift
@@ -1,0 +1,40 @@
+import SwiftUI
+
+public enum StatusAction : String, CaseIterable {
+  case none, reply, quote, boost, favorite, bookmark
+  
+  @MainActor
+  public var displayName: LocalizedStringKey {
+    switch self {
+    case .none:
+      return "settings.swipeactions.status.action.none"
+    case .reply:
+      return "settings.swipeactions.status.action.reply"
+    case .quote:
+      return "settings.swipeactions.status.action.quote"
+    case .boost:
+      return "settings.swipeactions.status.action.boost"
+    case .favorite:
+      return "settings.swipeactions.status.action.favorite"
+    case .bookmark:
+      return "settings.swipeactions.status.action.bookmark"
+    }
+  }
+  
+  public func iconName(isReblogged: Bool = false, isFavorited: Bool = false, isBookmarked: Bool = false)-> String {
+    switch self {
+    case .none:
+      return "slash.circle"
+    case .reply:
+      return "arrowshape.turn.up.left"
+    case .quote:
+      return "quote.bubble"
+    case .boost:
+      return isReblogged ? "arrow.left.arrow.right.circle.fill" : "arrow.left.arrow.right.circle"
+    case .favorite:
+      return isFavorited ? "star.fill" : "star"
+    case .bookmark:
+      return isBookmarked ? "bookmark.fill" : "bookmark"
+    }
+  }
+}
--- a/Packages/Env/Sources/Env/StatusAction.swift
+++ b/Packages/Env/Sources/Env/StatusAction.swift
@@ -1,9 +1,11 @@
 import SwiftUI
 
-public enum StatusAction : String, CaseIterable {
-  case none, reply, quote, boost, favorite, bookmark
+public enum StatusAction : String, CaseIterable, Identifiable {
+  public var id: String {
+    "\(rawValue)"
+  }
+  case none, boost, reply, quote, favorite, bookmark
   
-  @MainActor
   public var displayName: LocalizedStringKey {
     switch self {
     case .none:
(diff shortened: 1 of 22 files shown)


sothawo:
Just noticed: I should the current hard coded settings as default values

EvilOne:
Here are some examples of how others handle this kind of customisation. I am just posting these here for inspiration.

### Fastmail application
IMG_3326

### ProtonMail
IMG_3325


### TickTick
IMG_3323

EvilOne:
As for swipe xxx outer/inner, perhaps we can go for the text that the FastMail example uses?

EvilOne (commented):
Could you please copy all the new `Localizable.strings` entries to the other languages? Otherwise, other languages get things like this.

CleanShot 2023-02-11 at 08 33 51@2x

Packages/Env/Sources/Env/StatusAction.swift
```diff
@@ -0,0 +1,40 @@
+import SwiftUI
+
+public enum StatusAction : String, CaseIterable {
+  case none, reply, quote, boost, favorite, bookmark
+  
+  @MainActor
```
EvilOne, line 6: You should be able to remove this safely.

Packages/Status/Sources/Status/Row/StatusRowView.swift
```diff
@@ -453,43 +453,88 @@ public struct StatusRowView: View {
...
+      makeSwipeButton(action: preferences.swipeActionsStatusRight1)
+        .tint(theme.tintColor)
+    }
+    if preferences.swipeActionsStatusRight2 != StatusAction.none {
+      makeSwipeButton(action: preferences.swipeActionsStatusRight2)
+        .tint(.gray)
+    }
+  }
+
+  @ViewBuilder
```
EvilOne, line 467: Since we only have a single view being returned here, we can safely drop `@ViewBuilder`.

Packages/Status/Sources/Status/Row/StatusRowView.swift
```diff
@@ -453,43 +453,88 @@ public struct StatusRowView: View {
...
+      makeSwipeButton(action: preferences.swipeActionsStatusLeft1)
+        .tint(theme.tintColor)
+    }
+    if preferences.swipeActionsStatusLeft2 != StatusAction.none {
+      makeSwipeButton(action: preferences.swipeActionsStatusLeft2)
+        .tint(.gray)
+    }
+  }
+  
+  @ViewBuilder
```
EvilOne, line 479: Since we only have a single view being returned here, we can safely drop `@ViewBuilder`.

EvilOne:
Another future feature is not allowing users to select the same swipe gesture multiple times. For instance, I can do this today:

CleanShot 2023-02-11 at 08 45 11@2x

Sorry about the rambling here.

sothawo:
> As for swipe xxx outer/inner, perhaps we can go for the text that the FastMail example uses?

Like this one the best from the examples

sothawo:
> Could you please copy all the new `Localizable.strings` entries to the other languages? Otherwise, other languages get things like this.
> 
> CleanShot 2023-02-11 at 08 33 51@2x

I know, as I wrote, I wanted to wait until the PR is alright

Packages/Status/Sources/Status/Row/StatusRowView.swift
sothawo: If I remove these I seem to have to add explicit return statements which then lead to 
image

sothawo:
btw is there a way to have the "short" action, the "inner" one to be triggered by just swiping out the short way?

Dimillian:
I don't think so. Love this ^

seanlifetime:
> btw is there a way to have the "short" action, the "inner" one to be triggered by just swiping out the short way?
> 
> 
> 
I don't think it is possible using the stock controls.

seanlifetime:
Apple doesn't mention it anywhere on their page:

https://developer.apple.com/documentation/swiftui/view/swipeactions(edge:allowsfullswipe:content:)

sothawo:
I reworked the setting and now have this solution:

image

Seems to be more understandable

Dimillian:
This is really good! Making a few fixes and merging!

EvilOne:
Like the screen! Nice job!

sothawo:
thanks!

PR #787, opened 2023-02-11
Remove "iPhone" and "iPad" from translatable strings

te6-in:
I actually don't think `iPhone` and `iPad` should be translatable.

This removes `settings.display.section.phone` and `settings.display.section.ipad` from `Localizable.strings` and replace them with hard-coded strings.


Dimillian:
I agree

PR #799, opened 2023-02-11
Translated Swipe Actions to German

dertuxmalwieder:
Busy weekend, huh? ;-)

### DIFF
--- a/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
@@ -158,21 +158,21 @@
 
 "settings.display.show-ipad-column" = "Zweite Spalte aktivieren";
 
-"settings.general.swipeactions" = "Swipe Actions";
-"settings.swipeactions.navigation-title" = "Swipe Actions";
-"settings.swipeactions.status.action.bookmark" = "Bookmark";
-"settings.swipeactions.status.action.boost" = "Boost";
-"settings.swipeactions.status.action.favorite" = "Favorite";
-"settings.swipeactions.status.action.none" = "None";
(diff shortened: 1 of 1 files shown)


IceCubesApp/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -158,21 +158,21 @@
...
+"settings.swipeactions.status.action.bookmark" = "Lesezeichen setzen";
+"settings.swipeactions.status.action.boost" = "Boosten";
+"settings.swipeactions.status.action.favorite" = "Favorisieren";
+"settings.swipeactions.status.action.none" = "keine";
+"settings.swipeactions.status.action.quote" = "Zitieren";
+"settings.swipeactions.status.action.reply" = "Antworten";
+"settings.swipeactions.status.leading" = "Nach rechts wischen";
+"settings.swipeactions.status.leading.left" = "links";
+"settings.swipeactions.status.leading.right" = "links";
+"settings.swipeactions.status.trailing" = " Nach links wischen";
```
te6-in, line 172: I think the leading space is unnecessary. (The English one needs to be fixed too.)

dertuxmalwieder: Not sure about that ... I won't remove spaces without being sure that they're not intended...?

te6-in: Yeah, I think they are not intended and should be removed.

With a space:
SCR-20230212-it5

Without:
SCR-20230212-iuc

Maybe it'll be okay to remove it when the original English string gets fixed.

dertuxmalwieder: Will do … ;-)

Issue #747, opened 2023-02-09
Bug: Translation use the DNS domain of the poster, not the language of the toot...


jaco60:
## Environment:

 - OS: 
IOS 16.3
 - IceCubesApp version: 
 IceCubeApp v1.4.4

## Description

If a german (.de) poster toots in english, Icecube writes (in french) : "Traduire de allemand"

## Related Issues

Very bad french localisation : "traduire de allemand" is not french (as most of french localisation strings of the App...). "Traduire" would be enough

Dimillian:
Oué mais le français c'est compliqué roh

jaco60:
Raison de plus pour ne pas faire de grandes phrases :)

sothawo:
The question is: did the german user that posted in english had the language of the post set to english as well? Because the selected language for posting is sent along with the post data, so if this is set to german (which might be set as default for that user) then you'll see that behaviour

Havhingstor:
I think the language was probably wrongly set. I've seen examples where different posts from one user (https://mstdn.social/@stux) show the ability to translate from dutch or english, depending on the language of the post, suggesting that it only regards the language the user said the post is written.

jaco60:
I suppose it didn't... I've seen the same problem with a finnish user who admit to forget to switch to english before posting... But that's rather cumbersome to switch before a post, isn't it? Why not stay with a "one size fits all" message, as "Translate" ?

sothawo:
agreed for that link. 

But - and that has nothing to do with this issue - I find it important to always set the language for a post, because you can set in the settings on the server the languages you are interested in. For me I have set that to english, german and french. But I see posts from people in russian, where these are offered to be translated from english because the poster did not set the language accordingly.

jaco60:
Agree too... It is important to indicate the language of the post. But, if this is not done, no need to deceive the reader even more :)

Dimillian:
Everyone is bugging me about this, I don't know what to do. At this point I'm considering not using the post lang but running on device ML thing to detect the lang and use that as the true post language.

sothawo:
Having something to detect the language might probably be the best solution. 

The API definition btw does not require language when posting a status; I just tested the server's behaviour. I sent in a status without a language and got back "de" (default setting for my profile). Quite interesting that the mastodon server in this case returns the default language although I did not set any and the field in the returned status is nullable as well.

So if you got something in mind to detect the language on the device, I wouldn't change the current behaviour now.

[...]

PR #800, opened 2023-02-11
Automatically detect language of posts, improve language detection when posting

Havhingstor:
The target-language for the translation option is automatically determined via Apple's NaturalLanguage-Library. The detection works similar to the already implemented detection when posting, but hashtags, emojis (*:emoji:*) and "@"-links are filtered out, so that the processed text is cleaner. If the detected language is wrong the user still has the option to translate from the transmitted (*"secondary"*) language in the context menu.
The language-detection when posting is also reworked: The language is always detected, even when the user explicitly selected another one. The user is however always asked in which language he wants to post, since quietly changing the language before posting can confuse.

That should address issue #747 at least to a certain degree.

Optionally I could also see a scenario where the translation from the secondary language is also accessible via a button in main view (not only the context-menu). That might be easier to understand, depending on the accuracy of the language detection.

### DIFF
--- a/IceCubesApp/Resources/Localization/ca.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/ca.lproj/Localizable.strings
@@ -360,6 +360,8 @@
 "status.editor.error.upload" = "Error en la pujada";
 "status.editor.language-select.navigation-title" = "Selecciona la llengua";
 "status.editor.language-select.recently-used" = "Recents";
+"status.editor.language-select.confirmation.detected-%@" = "Post in %@ (Detected language)";
+"status.editor.language-select.confirmation.selected-%@" = "Post in %@ (Selected language)";
 "status.editor.media.edit-image" = "Edita la imatge";
 "status.editor.media.image-description" = "Descripció de la imatge";
 "status.editor.mode.edit" = "Edita la publicació";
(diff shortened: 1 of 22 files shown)


Havhingstor:
It might also be possible to find a better replacement for "secondary" language, maybe "transmitted" or "set" language

nekolaboratory:
May be "Secondary language" might be rephrased as "other candidates".

Dimillian:
This is really good for the editor part, for the status part I'm not sure we want to take in the performances hit it brings.
Processing the string and detecting the language as you scroll the timeline is heavy. 
It would need to be async for it to be safer, but if we do that, then it means it'll update the button once it's done processing, and this would also brings lag with SwiftUI. The lazy var you have don't really protect anything as it's called while dequeing / scrolling statuses.

I recommend removing the part in the status row and keeping the part with the langage detection in the editor, because this is really great.

Dimillian:
I've changed a bit the code on main: 

- It now prefer to use the Mastodon API when available
- I'm not forcing the lang of the post to be different from the user lang (local or server) anymore to show the translate option in the ... menu. 

I think your change are great, but you can drop the one in the status, let just have the confirmation from the editor before posting.

Havhingstor:
I've removed the language-detection on the timeline for now. I might come back to it later on, but for now I've focused on the editor.

PR #801, opened 2023-02-11
Update Japanese Localizable.strings

nekolaboratory:
* Swipe actions

### DIFF
--- a/IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
@@ -161,21 +161,21 @@
 
 "settings.display.show-ipad-column" = "セカンドカラムを有効にする";
 
-"settings.general.swipeactions" = "Swipe Actions";
-"settings.swipeactions.navigation-title" = "Swipe Actions";
-"settings.swipeactions.status.action.bookmark" = "Bookmark";
-"settings.swipeactions.status.action.boost" = "Boost";
-"settings.swipeactions.status.action.favorite" = "Favorite";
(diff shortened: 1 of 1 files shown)


IceCubesApp/Resources/Localization/ja.lproj/Localizable.strings
```diff
@@ -161,21 +161,21 @@
...
+"settings.swipeactions.status.action.bookmark" = "ブックマーク";
+"settings.swipeactions.status.action.boost" = "ブースト";
+"settings.swipeactions.status.action.favorite" = "お気に入り";
+"settings.swipeactions.status.action.none" = "何もしない";
+"settings.swipeactions.status.action.quote" = "引用";
+"settings.swipeactions.status.action.reply" = "返信";
+"settings.swipeactions.status.leading" = "右にスワイプ";
+"settings.swipeactions.status.leading.left" = "左";
+"settings.swipeactions.status.leading.right" = "右";
+"settings.swipeactions.status.trailing" = " 左にスワイプ";
```
te6-in, line 175: I think the leading space is unnecessary. (The English one needs to be fixed too.)

nekolaboratory: Fixing only this language first would be inconsistent, so I'll leave it as is until later when I fix the English version, which is the source of the branch.

Thank you for your advice!

nekolaboratory:
@Dimillian Why did you close my PR?

Dimillian:
Because another translation was merged.

nekolaboratory:
Yes. It was committed later than mine.

And some translations are wrong.

PR #807, opened 2023-02-12
Fix AppAccountView touch area

te6-in:
This fixes the spacer not being tappable in `AppAccountView`.

SCR-20230212-jxm


Dimillian:
Thanks!

PR #808, opened 2023-02-12
Improve swipe actions and settings

te6-in:
This removes an unnecessary leading space from `settings.swipeactions.status.trailing`.
Definitely not a big deal, but details matter 🙂

With a space:
SCR-20230212-it5

Removed:
SCR-20230212-iuc

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusRowView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowView.swift
@@ -520,7 +520,7 @@ public struct StatusRowView: View {
     } label: {
       Text(action.displayName)
       Image(systemName: action.iconName(isReblogged: viewModel.isReblogged, isFavorited: viewModel.isFavorited, isBookmarked: viewModel.isBookmarked))
-        .foregroundColor(.red)
+        .environment(\.symbolVariants, .none)
     }
   }
   
@@ -534,6 +534,7 @@ public struct StatusRowView: View {
     } label: {
       Text(action.displayName)
       Image(systemName: action.iconName(isReblogged: viewModel.isReblogged, isFavorited: viewModel.isFavorited, isBookmarked: viewModel.isBookmarked))
+        .environment(\.symbolVariants, .none)
     }
   }
 }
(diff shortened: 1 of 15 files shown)


te6-in:
Updates:

- Swipe actions settings
  - SCR-20230212-mvq
  - Make the settings view be consistent with other views
  - Add icons to menus
  - Separate the `none` option into a different section
  - Reorder options to make them consistent with inline status buttons

- Actual swipe buttons
  - SCR-20230212-nkc
  - Show hollow / filled icons and different strings depending on `isReblogged`, `isFavorited`, `isBookmarked`
  - Show different button colors based on the action

Packages/Status/Sources/Status/Row/StatusRowView.swift
```diff
@@ -455,23 +455,35 @@ public struct StatusRowView: View {
...
-        .tint(theme.tintColor)
+        .tint(
+          preferences.swipeActionsStatusTrailingRight == .boost ? theme.tintColor :
+            preferences.swipeActionsStatusTrailingRight == .favorite ? .yellow : preferences.swipeActionsStatusTrailingRight == .bookmark ? .pink : .gray
+        )
     }
     if preferences.swipeActionsStatusTrailingLeft != StatusAction.none {
       makeSwipeButton(action: preferences.swipeActionsStatusTrailingLeft)
-        .tint(.gray)
+        .tint(
```
Dimillian, line 465: This should be refactored in some function to make it shorter and easier to read.

te6-in: @Dimillian Please take a look :)

sothawo:
The settings now show the icon in front of the text I left the icons out at this place, because it makes the right side looking misaligned.

image

And as for the colors: I had the outer swipes in the tint color and the inner ones in gray. After this change I get a big yellow swipe for the favorite (which does not fit my color scheme at all), the boost is in the tint color, and on my right side the quote and reply have the same color and I cannot see where one ends and the other starts

image

image

te6-in:
@sothawo Could you take a look at #825 and tell me what you think? About the icons, I too was thinking that they were somewhat excessive and looked clunky in the menu.

PR #810, opened 2023-02-12
Update Simplified Chinese translations

Jerry23011:
### Changes
- Add translation for Custom Emojis
- Add translations for Accessibility
- Add translations for Swipe actions settings
### References
[Apple Support](https://support.apple.com/zh-cn/HT208204)

@nixzhu would you like to take a look :)


nixzhu:
LGTM

PR #815, opened 2023-02-12
Smol fixes for Basque

xabirequejo:
Also, string "account.detail.n-fields %lld" doesn't seem to use localisation on 1.4.6. Will create an issue.


te6-in:
I looked up the Localizable.strings file and it seems that the entire line is missing for both. Doesn't look like a typo issue for me.

PR #825, opened 2023-02-13
Improve swipe actions

te6-in:
[context](https://github.com/Dimillian/IceCubesApp/pull/808#issuecomment-1427116198)

- Remove icons from menus
- Fix two gray buttons together being difficult to distinguish
- Add a new toggle to use theme tint colors instead of button colors (of boost, favorite, bookmark ones)
  - With the toggle turned off (use default button colors):
  - SCR-20230213-hwe
  - on (use theme colors):
  - SCR-20230213-hw4

What I'm not sure about
- Where should be the toggle at? display settings or swipe actions settings?
- Default value

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/SwipeActionsSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/SwipeActionsSettingsView.swift
@@ -59,6 +59,9 @@ struct SwipeActionsSettingsView: View {
             }
           }
         }
+        Toggle(isOn: $userPreferences.swipeActionsUseThemeColor) {
+          Label("Use theme colors instead of default colors", systemImage: "paintbrush.pointed")
+        }
       }
       .listRowBackground(theme.primaryBackgroundColor)
     }
--- a/IceCubesApp/App/Tabs/Settings/SwipeActionsSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/SwipeActionsSettingsView.swift
@@ -62,6 +62,11 @@ struct SwipeActionsSettingsView: View {
         Toggle(isOn: $userPreferences.swipeActionsUseThemeColor) {
           Label("settings.swipeactions.status.use-theme-colors", systemImage: "paintbrush.pointed")
         }
+        Picker(selection: $userPreferences.swipeActionsIconStyle, label: Label("Icon style", systemImage: "text.below.photo")) {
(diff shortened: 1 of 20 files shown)


Dimillian:
I'll wait for @sothawo opinion but here is mine:

* I like colors related to the actions, so I would be ok with the default settings being button colors sound good to me.
* The toggle should be alongside the swipe actions settings.

sothawo:
2 places we are talking about:

1. the icons in the picker in the settings pane (not in the popup, there the alignment is alright), but where the icons are shown in front of the label: They should be removed from the settings pane but be kept in the popup. When I implemented it  - before that change came in that added these icons -  I deliberately left the icons out because I thought that looks not good.

2. the actual swipes: 
 - I'd rather keep the icons instead of text only, but this could be a preference setting as well (icons/text/icons+text)
 - the colors being configurable (like it is now as default) or using tint as the outer and grey as the inner. It's hard to decide if there are two grey buttons side by side, as they can be randomly configured to be on either side so you do not know if the yellow favorite is next to the tinted boost and the gray reply and quote on the other side, or the yellow favorite is togeher with a gray reply and the tinted boost with another gray. So for me the two options would be
   1. tint/gray on both sides like I had it in the first version
   2. icon colors like now

Dimillian:
Oh yes. We definitely need icons in the swipe actions.

te6-in:
In the case of the settings view, this PR completely removes the icons from the settings. I'd like to make the icons appear only when the picker is tapped and the context menu is open, but I couldn't find out how to do so.

Also, I'll add a new picker for showing only icons or icons with text.

In terms of the colors, I think I'll keep the default value to false (color matches its action), but let me know if have any ideas.

sothawo:
I can have a look when I'm back from work later this evening

sothawo:
Looks fine for me, love it that I can keep my preferred coloring 😄

Dimillian:
LGTM?

te6-in:
Added the icon style picker. I think this is good to go? 🙂

sothawo:
There I take 15 minutes for breakfast and when coming back to the laptop, the PR is merged. But the only thing I wanted to say is that perhaps the option "text only" could have been added. But I wouldn't use that anyway.

te6-in:
@sothawo Actually I didn't really think many people will use the text only option. Maybe we can add it later when there is a request. 😄

PR #832, opened 2023-02-13
Fixed non-existing Json key decoding error

nekolaboratory:
The relationship endpoint of the instance to which I belong is as follows, and the notification key does not exist.

```
   {
      "id":"1",
      "following":true,
      "showing_reblogs":true,
      "followed_by":false,
      "blocking":false,
      "blocked_by":false,
      "muting":false,
      "muting_notifications":false,
      "requested":false,
      "domain_blocking":false,
      "endorsed":false,
      "note":""
   }
```

Decoding this will result in an error.
So I tried decoding while checking if the key exists.

### DIFF
--- a/Packages/Models/Sources/Models/Relationship.swift
+++ b/Packages/Models/Sources/Models/Relationship.swift
@@ -31,3 +31,22 @@ public struct Relationship: Codable {
           notifying: false)
   }
 }
+  
+public extension Relationship {
+  init(from decoder: Decoder) throws {
+    let values = try decoder.container(keyedBy: CodingKeys.self)
+    id = try values.decodeIfPresent(String.self, forKey: .id) ?? ""
+    following = try values.decodeIfPresent(Bool.self, forKey: .following) ?? false
+    showingReblogs = try values.decodeIfPresent(Bool.self, forKey: .showingReblogs) ?? false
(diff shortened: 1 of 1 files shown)


sothawo:
Interesting, the [API documentation](https://docs.joinmastodon.org/entities/Relationship/#notifying) does not show this as nullable.

Dimillian:
Not sure how I want to handle it. I don't like custom decoder... but I don't like making bool values optional.

sothawo:
What about having a model with nullable property to be used in decoding and then the "real" model that can be initialized from the first one that will set defaults on missing values?

Dimillian:
Then I think I would prefer the custom decoder solution provided here :p

nekolaboratory:
> Interesting, the [API documentation](https://docs.joinmastodon.org/entities/Relationship/#notifying) does not show this as nullable.

The server version of my instance was 3.2.1.
According to the API documentation, the notify key was added in 3.3.0.

nekolaboratory:
@Dimillian

As mentioned above, unless the version by the server is the latest, a decode error will occur because the key does not exist.

In terms of display, when I open the instance I belong to in the App, the follow list page remains loading and does not progress.

In order to deal with this problem and eliminate the version difference between servers, I would like to check the existence of the key before decoding. How about my fix?

PR #833, opened 2023-02-13
Use previously unused LocalizedStringKey in language selector

icod:
The string "Recently Used" was not replaced by a `LocalizedStringKey` yet. 

image

I checked and the string "status.editor.language-select.recently-used" is available in all the present localizations.


icod:
It now uses the readily available localized strings:

Scherm­afbeelding 2023-02-13 om 15 21 08

PR #843, opened 2023-02-14
Use semver library for version checks

joebarbere:
This could be useful if there is a need in the future to check against more complex version strings.

i.e. 5.0.0-beta

Uses https://github.com/sersoft-gmbh/semver


Dimillian:
Had to revert, it doesn't work for 4.1.0rc3 which is mastodon.social version for example.

PR #856, opened 2023-02-14
Ignore the report button on my post

nekolaboratory:
Fixed the report button in the status action menu to be hidden for my posts.


Dimillian:
But that on purpose, why would you report your own post? Your own post have a different set of options.

nekolaboratory:
Sorry, there was a mistake in the description.

In the latest main branch, the report button is also displayed on my post.

PR #878, opened 2023-02-15
Add account name as navigation header in profile view

### LINKED ISSUE #872: Bug: No header text in scrolled Profile view?
## Environment:

 - OS: iOS 16.3
 
 - IceCubesApp version: 1.4.8
 

## Description

A very tiny inconvenience, but...  When you send IceCubes to background while scrolled down in the Profile view, and come back hours later, there is no obvious clue where you are, at least for color-challenged people: 

iPhone-2023 02 14-10 04 18 000 when you left ice scrolled down your profile

Could there be something written in the blank header space? Like there is for all the other bottom button views? 

## Related Issues

 - [ ] Search that this bugs don't already exist before creating it.

sothawo:
Closes #872

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailView.swift
+++ b/Packages/Account/Sources/Account/AccountDetailView.swift
@@ -127,6 +127,7 @@ public struct AccountDetailView: View {
       FiltersListView()
     })
     .edgesIgnoringSafeArea(.top)
+    .navigationTitle(viewModel.account?.displayName ?? "")
     .navigationBarTitleDisplayMode(.inline)
     .toolbar {
       toolbarContent


Dimillian:
Buuuut it look bad if you didn't scrolled right? I've removed it because of that. Since it's a List I can't get the scroll offset and show it accordingly anymore :/
I was thinking to show the nav bar at all time on this screen

sothawo:
It's looks ugly above the image, I admit. as the first idea I had to fix the issue. So no problem if you don't want to merge it.

Dimillian:
Closing for now, will think of a different solution :)

PR #905, opened 2023-02-16
Update Simplified Chinese translations

Jerry23011:
### Changes
- Add new translations
- Refined some previous translations

@nixzhu could you check it out :)


nixzhu:
LGTM

Issue #829, opened 2023-02-13
Bug: notifications do not refresh when you open the app and select the notifications tab


markrprior:
## Environment:

 - OS: 
 iOS 16.1.2
 - IceCubesApp version: 
 IceCubesApp v1.4.7

## Description

If you open the app and immediately select the notifications tab it doesn't update.

![IMG_4423](https://user-images.githubusercontent.com/6236685/218432922-3dfbd82d-b204-4cf5-aef0-7c9510eadc82.PNG)

## Related Issues

 - [X] Search that this bugs don't already exist before creating it.

PR #917, opened 2023-02-17
Fix leak memory (Thread 1: EXC_BAD_ACCESS (code=EXC_I386_GPFLT))

nekolaboratory:
Closes: #829

### DIFF
--- a/Packages/Models/Sources/Models/ConsolidatedNotification.swift
+++ b/Packages/Models/Sources/Models/ConsolidatedNotification.swift
@@ -26,7 +26,7 @@ public struct ConsolidatedNotification: Identifiable {
     self.type = type
     self.createdAt = createdAt
     self.accounts = accounts
-    self.status = status
+    self.status = status ?? nil
   }
 
   public static func placeholder() -> ConsolidatedNotification {


Dimillian:
Are you sure it did anything?

nekolaboratory:
Yes.
The problem is exactly the same as #829, and it happens until you manually reload.

I found the reason for this problem when I debugged.

https://github.com/Dimillian/IceCubesApp/blob/5caa0b647879a3a34a3860f51eb413ab98162aab/Packages/Notifications/Sources/Notifications/NotificationRowView.swift#L110

https://github.com/Dimillian/IceCubesApp/blob/5caa0b647879a3a34a3860f51eb413ab98162aab/Packages/Notifications/Sources/Notifications/NotificationRowView.swift#L134

When you access notification.status in the above location, "(Thread 1: EXC_BAD_ACCESS (code=EXC_I386_GPFLT))" occurs when you open the notification screen for the first time, causing an infinite loop.

This seemingly meaningless syntax solved that problem.

Dimillian:
Are you sure an Xcode clean will not solve the issue? I've not been able to reproduce this crash and I see no report of it in the reporting.

nekolaboratory:
We have confirmed that there is a bug in the official release version of IceCubesApp v1.5.1 installed from the App Store.

A cleanup of Xcode also reproduces the issue on the latest main branch.

By the way, the instance I checked is "best-friends.chat".

Dimillian:
Well, it's crazy but ok.

PR #924, opened 2023-02-18
Update to Nuke 12 (Beta 1)

kean:
It contains a couple of performance improvements:

- Use native `SwiftUI.Image` for all rendering
- Reduce the amount of fields in `LazyImage`
- Reduce the number of `body` reload of `LazyImage` by extracting progress updates to a separate observable object

More info here: https://github.com/kean/Nuke/releases/tag/12.0.0-beta.1

**Warning**: I haven't tested some screens. It says beta 1, but it's a (relatively) safe change that only affects `LazyImage` and for the most part it's only code removal.


Dimillian:
Fixed few places but let's go!

PR #925, opened 2023-02-18
Status as class (performance improvement)

kean:
It's inefficient to use structs for types that contain a lot of fields, especially if they are either reference types or structs that use references internally (e.g. String and its underlying CoW representation). Every time you pass a struct around, it performs a retain/release for every property. By changing it to a class, there is only one retain/release call needed. And there is much less memory needed to copy these values around – only the pointer itself. Reference: https://developer.apple.com/videos/play/wwdc2016/416/

**Before**

Screenshot 2023-02-17 at 8 38 45 PM

**After**

Screenshot 2023-02-17 at 8 42 37 PM

### DIFF
--- a/Packages/Models/Sources/Models/Account.swift
+++ b/Packages/Models/Sources/Models/Account.swift
@@ -1,6 +1,10 @@
 import Foundation
 
-public struct Account: Codable, Identifiable, Equatable, Hashable {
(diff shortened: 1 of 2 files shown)


Dimillian:
I was hesitant to do that because... well I think it's sad. But necessary yes. Too bad we can't have only clean struct. I've also made Account a class.

PR #926, opened 2023-02-18
Update Simplified Chinese Translation

DejavuMoe:


Dimillian:
@Jerry23011 @nixzhu If you're able to check this :)

Jerry23011:
It's literally a reasonable change since it matches the translations on Mastodon Web.
However, I personally don't quite agree with whoever did that translation (for Mastodon Web) because they're using the one on the Bird app, while the bird app uses "like", Mastodon uses "favorite".
Apple's translation for "favorite" in their own app is "收藏" (the original one), and that's how I made the decision in the first place.
Nevertheless, those changes look good to me👍🏻. Probably going to see what I can do with the official translations later..

nixzhu:
LGTM

Jerry23011:
Follow-up: The Chinese community has already reached consensus on the translation of `Favorite`. See [this post](https://bgme.me/@mastodon_chinese_translate/108290374067275641)

PR #927, opened 2023-02-18
Use semver library for version checks - attempt #2

joebarbere:
So I originally used https://github.com/sersoft-gmbh/semver but as you know it failed to parse 4.1.0rc3. This was due to the fact that Mastodon does not actually fully conform to the semver spec for their version numbers.

This time around I've used a different library https://github.com/ddddxxx/Semver which had a more friendly license. I updated the regex so that Mastodon versions will work and added a few unit tests.

Please merge if you feel it will be useful. No worries if not though. I had fun learning how to setup unit tests in Xcode, so its a win for me 😄

### DIFF
--- a/Packages/Env/Package.swift
+++ b/Packages/Env/Package.swift
@@ -29,5 +29,9 @@ let package = Package(
         .product(name: "KeychainSwift", package: "keychain-swift"),
       ]
     ),
+    .testTarget(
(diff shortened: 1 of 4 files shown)


Dimillian:
Will stay with our simple solution for now, until I have more time too look at it. But thanks!

PR #929, opened 2023-02-18
Display and edit notes on profiles

### LINKED ISSUE #165: Feature request: create, edit and view private note for users
Have the option to create, edit and view a private note for user accounts.

sothawo:
This PR adds displaying and editing of private notes on profiles. 

## Displaying

I choose to display them above the normal profile description, as these notes were written by the user looking at them, not by the person whose profile we're looking at. So for me they are more important and should be on top:

grafik

## Editing

In order to be able to edit, I added a new menu entry:

grafik

which will bring up this sheet:

grafik

In order to delete the note, it's just necessary to remove all text, same as in the web UI

grafik

## misc

When the note is shorter than one line, it is displayed centered, I did not find a way to fix that, as soon as there a more lines, they are .leading aligned.

grafik

Closes #165

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailHeaderView.swift
+++ b/Packages/Account/Sources/Account/AccountDetailHeaderView.swift
@@ -150,6 +150,12 @@ struct AccountDetailHeaderView: View {
           }
         }
       }
+      
+      if let note = viewModel.relationship?.note, !note.isEmpty,
+          !viewModel.isCurrentUser {
+        makeNoteView(note)
+      }
+
       EmojiTextApp(account.note, emojis: account.emojis)
         .font(.scaledBody)
         .padding(.top, 8)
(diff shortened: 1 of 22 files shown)


Packages/Account/Sources/Account/AccountDetailHeaderView.swift
```diff
@@ -193,6 +199,23 @@ struct AccountDetailHeaderView: View {
...
     }
   }
+  
+  @ViewBuilder
+  private func makeNoteView(_ note: String) -> some View {
+    VStack(alignment: .leading, spacing: 4) {
+      Text("account.relation.note.label")
+        .foregroundColor(.gray)
+      Text(note)
+        .frame(maxWidth: .infinity)
```
EvilOne, line 209: Could you add alignment: .leading to fix the one line issue? Not by a computer to confirm if it works.

sothawo: The `Text` has no `.alignment`, and on the `VStack` containing the text it is already set.

Dimillian:
This is really nice!!

PR #930, opened 2023-02-18
Add image prefetch to the timeline + resize close #931

### LINKED ISSUE #931: Image size in the timeline
I noticed that the app uses `url` and not `previewUrl` for images in the timeline. Is there a good reason not to use the previews?

Let's take the following image from the timeline as an example:

- Original (1247x1663) https://files.mastodon.social/media_attachments/files/109/884/136/729/763/293/original/1ee92b4fcac2dabc.jpeg
- Preview (416x554) https://files.mastodon.social/media_attachments/files/109/884/136/729/763/293/small/1ee92b4fcac2dabc.jpeg

The original is too high resolution for the timeline. The main concern isn't necessary the image data size, but how much memory it takes to bitmap it: ~8 MB. This will lead to high memory usage.

The image previews displayed in a grid take only 172 pt width. 416px is plenty to display it with high quality and in comparison, it takes less than 1 MB to bitmap. If there is only one image, it'll might make sense to use the original, but scale it.

I suggest the following improvements:
- Use `previewUrl` (note: [prefetcher](https://github.com/Dimillian/IceCubesApp/pull/930) needs to use the same URLs)
- Use `ImageProcessors.resize` to resize the images in the background to exactly match the display size (challenge: computing this size ahead of time and communication it to the [prefetcher](https://github.com/Dimillian/IceCubesApp/pull/930) too)
- A further optimization is to crop the preview to reduce memory usage even further

kean:
The timeline now prefetches the images, so that they are ready to be displayed by the time user scrolls to them and they become visible.

### DIFF
--- a/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
+++ b/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
@@ -206,7 +206,7 @@ public struct StatusRowMediaPreviewView: View {
           switch type {
           case .image:
             ZStack(alignment: .bottomTrailing) {
-              LazyImage(url: attachment.url) { state in
+              LazyImage(url: attachment.previewUrl ?? attachment.url) { state in
                 if let image = state.imageContainer?.image {
                   SwiftUI.Image(uiImage: image)
                     .resizable()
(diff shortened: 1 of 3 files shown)


kean:
Update spacing to 2 .

PR #937, opened 2023-02-18
Redesigned swipe actions settings

icod:
This redesigns the Swipe Actions settings screen.

| Before | After |
| --- | --- | 
| Simulator Screen Shot - iPhone 14 Pro - 2023-02-18 at 19 52 21 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-18 at 20 03 47 | 

## Changes

It changes the hierarchy of the settings screen from **general** swipe actions to **post** swipe actions. If, in the future, other entities need their own swipe settings, they should get their own separate settings view. So that from the 'Swipe Actions' screen the user can navigate to 'Post Swipe Actions' and 'xx Swipe Actions'.

Secondly, it changes the concept of left/right actions on both sides of the post to primary/secondary. In my opinion this is better suited to the way these actions are applied to a given post: when the user swipes _all the way_ to the left, the right-hand-side primary action is applied; when the user swipes all the way to the right, the left-hand-side primary action is applied. I hope this causes less confusion to the user. To fully follow this concept of primary/secondary, the picker for the secondary action is disabled and set to 'None', if the users sets the primary action to 'None'.

Third, the 'Use Theme Colors' label has been abbreviated and now has a little explanatory text below it. (Which is for now the (long) text that used to be the label text.)

## To Do

If this is indeed an improvement over the current screen, I will add the localization strings and clean up the TODO remarks.

Also, maybe the navigation title should change from 'Swipe Actions' to 'Post Swipe Actions' or something like that.

### DIFF
--- a/IceCubesApp.xcodeproj/project.pbxproj
+++ b/IceCubesApp.xcodeproj/project.pbxproj
@@ -919,7 +919,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 1.5.6;
+				MARKETING_VERSION = 1.5.7;
 				PRODUCT_BUNDLE_IDENTIFIER = "$(BUNDLE_ID_PREFIX).IceCubesApp.IceCubesNotifications";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
@@ -949,7 +949,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 1.5.6;
+				MARKETING_VERSION = 1.5.7;
 				PRODUCT_BUNDLE_IDENTIFIER = "$(BUNDLE_ID_PREFIX).IceCubesApp.IceCubesNotifications";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
@@ -980,7 +980,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 1.5.6;
+				MARKETING_VERSION = 1.5.7;
 				PRODUCT_BUNDLE_IDENTIFIER = "$(BUNDLE_ID_PREFIX).IceCubesApp.IceCubesShareExtension";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
@@ -1010,7 +1010,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 1.5.6;
+				MARKETING_VERSION = 1.5.7;
 				PRODUCT_BUNDLE_IDENTIFIER = "$(BUNDLE_ID_PREFIX).IceCubesApp.IceCubesShareExtension";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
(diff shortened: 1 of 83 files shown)


Dimillian:
@sothawo what do you think? I like it this is easier to read IMO.

sothawo:
Not sure about that.

First, splitting this into different sections means that if there are more swipe actions in the future, we need different setting sheets for them as you wrote (and I hope we will have more; I'd like to be able to swipe on the top across the name of the displayed timeline to switch to the next one, from home to local to federated etc, and then there should be a section in the configuration where to set which timelines might be included in that swipe list), but currently we can stick to this single one

The second thing is, I know what the primary (leading) is and what the secondary, as you do and probably every SwiftUI developer. But does the **user** of the app know what primary and secondary mean in this context? That's why I put in the icons after the discussion on my first PR for this. I am not really happy with these icons, but they were the best I found in the icons collection. The current setting seems a little overloaded with icons.

But what if we add an explanatory text at the top? One that describes what a primary action is etc (smaller font, gray, like the one at the bottom)? This change would be fine for me if we have such an additional info text.

EvilOne:
Perhaps a little context qould help by reading the original PR #785.

icod:
Thanks guys, for the comments!

The original PR is very useful for context indeed (thanks @EvilOne ).  This concept of left/right and primary/secondary is complicated to convey to the user in a few stock controls. Maybe a little footer text will help, as suggested by @sothawo 

I’ll try some things later on and keep you posted. 

By the way, do we have consensus on the question whether all entities’ swipe actions should be in one view? Or separate views?

sothawo:
I think we should leave them in one view. Separating them will lead to a more complicated way to reach them. Either we would have to had entries like _swipe action for posts_, _swipe actions for something else_ in the main settings menu, making this longer, or we'd need to introduce a layer of selection by going from _swipe action_ to a view that lists the different ones and then goes on to the different views. If we have so many different swipe actions to configure that one view is not enough we can think about splitting it out.

So I'd prefer:
* keep one view for swipe action configuration and keep the ones for posts in a section as it is currently
* make the current view cleaner by removing the icons, rename to primary, secondary with an explanatory text above describing what these terms mean.

icod:
An update: 

- I have kept the swipe actions in one view, that is to say: the post swipe actions are back in one section. 
- I have made the view cleaner by renaming to primary/secondary and greying out the "Swipe Left/Right" a little. Also remove the icons. 

Not sure about the helper text at the top of the screen, though. 

| Text _above_;<br>_no_ padding on action pickers | Text _below_;<br>16-pt-padding on action pickers | Text _below_; <br>_no_ padding on action pickers | 
| --- | --- | --- |
| Simulator Screen Shot - iPhone 14 Pro - 2023-02-20 at 21 44 03 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-20 at 21 50 49 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-20 at 21 53 53 |

I think the third one is best. What do you think? 

I like the distinction between the "Post" section and the "Appearance" section. Do I understand correctly that the "Appearance" settings apply to all types of swipe actions (not only to those for posts)?

te6-in:
I also like the third one the most.

I think it's a nice approach to apply the same appearance preferences to future swipe actions so it'll be okay to separate those two and put them in the same page. But it'll be better to move the appearance section to the top when other sections for swipe settings (Account, Notification, etc...) come in later, in my opinion.

sothawo:
3rd one as well. Having the text above separates it from the post context.

As for the appearance, not every swipe action will have some visual appearance, think of a possible swipe action on the top bar when showing a timeline, this would switch the timeline but have no displayed text or icon. 
But if there will be other swipe actions that show an icon, I think that the colouring should be consistent, so having this out of the post section is fine for me.

Dimillian:
I like third one too. I think this is ready to merge?

icod:
No, I will have to do the localization strings first.

EvilOne:
The 3rd one indeed looks the best! Nice!

icod:
Added localizable strings, so I think this PR is ready, @Dimillian

(Had some merge conflicts after rebasing the branch. Sorry for the long commit list!)

PR #939, opened 2023-02-18
Added a tip that it's a video, and not an image when autoplay is off

Dean151:
**Motivation**:
When autoplay is off in the settings, images are not distinguishable from videos. To prevent missing fun or info from videos, we need tip them

**Proposed solution**:
Adding a "play" tip at the bottom left ; with the same stylish than "ALT"
The play tip only shows when autoplay is off.
The position have been chosen to not overlay with VideoPlayer standard controls that might appear sometime.

Screenshot of the actual look of the video tip

### DIFF
--- a/Packages/Status/Sources/Status/Media/VideoPlayerView.swift
+++ b/Packages/Status/Sources/Status/Media/VideoPlayerView.swift
@@ -1,4 +1,5 @@
 import AVKit
+import DesignSystem
 import Env
 import SwiftUI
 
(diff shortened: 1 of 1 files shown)


sothawo:
My personal preferences would be:
* center the overlay in the middle of the image, it might be missed in the corner
* make it a little larger

Dean151:
That was my first attempt, but as the viewer in the timeline is a full AVPlayer, sometime the controls would appear and overlay and we have little control over that with VideoPlayer from SwiftUI. Unless switching VideoPlayer with a Image, with a lot of work, or using a ViewRepresentable mecanism to embed a AVPlayerController, this is ugly.

Dean151:
I've made an update ; proposing a solution.

It's still overlaying with Player controls at some occasions, but since the "tip" is large enough, the play button is completely hidden behind the circle.

So this is what it looks like now:

new preview for video tip new preview with player control overlays src=

Dimillian:
Feel fine to me now. Until we have a better video player integration, it's ok to have that. I wish I could auto play video in QuickLook but it's not the case :/

sothawo:
I dont see the tip overlaying the video when playing (running on your checked out branch), when starting the video, it opens in a new window

Dean151:
> Feel fine to me now. Until we have a better video player integration, it's ok to have that. I wish I could auto play video in QuickLook but it's not the case :/

@dimillian I think it would be time to consider leaving Quicklook ; and starting to have something more custom with more control over media previews ; and full screen view.

PR #942, opened 2023-02-18
Optimize viewId

kean:
It was high on the list in Time Profiler. String interpolation and date formatting are expensive. This change eliminates both.


Dimillian:
I didn't knew it was still showing up. Nice!

PR #944, opened 2023-02-19
Updated EmojiText to 2.x

Dean151:
Interface used by IceCubeApp has not changed from 1.x to 2.x 

Looking at the library diff shows new features like SFSymbol powered emojis. 

Not like we're going to make use of that https://github.com/divadretlaw/EmojiText/compare/v1.2.0...2.0.2

### DIFF
--- a/IceCubesApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/IceCubesApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -23,8 +23,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/divadretlaw/EmojiText",
       "state" : {
-        "revision" : "3d8f6196de59634352ed317cf6ab4e292eeaaf44",
-        "version" : "1.2.0"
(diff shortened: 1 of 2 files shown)


divadretlaw:
LGTM. Since IceCubesApp already uses Nuke 12 (beta) I had to try if it breaks anything, but it doesn't seem to be the case at the moment, but that obviously depends on further Nuke 12 development.

In case some future beta update breaks anything, we can also just make a `nuke-12` branch on `EmojiText`.

The only way IceCubesApp is using the new SF Symbol Emoji is that the default placeholder is a SF Symbol Emoji now, instead of a LocalEmoji. But it should look more or less the same

Dimillian:
Wow, I didn't knew you made that much update @divadretlaw :D 
Do you think the performances are even better? need to dig into your code to see what changed.

Dimillian:
Seems all good!

kean:
Didn't realize the dependencies were also using Nuke :D

@divadretlaw, I looked through the code. `EmojiText` uses only the core framework `Nuke` and its async/await API. There were (almost) no changes in 12.0 to the main framework. The main changes are in `NukeUI`. It's safe to upgrade.

PR #945, opened 2023-02-19
Fix In-App Safari

divadretlaw:
## In-App Safari

The current `.fullScreen` way of presenting doesn't really feel native in my opinion since `SFSafariViewController` is supposed to be pushed in from the right and able to swipe back fluently.

### Current

Currently the swipe back "works" it just dismisses the entire sheet immediately.

https://user-images.githubusercontent.com/6899256/219949272-25cd598b-ff74-4745-846b-3ed710d46d7f.mov

### This PR

https://user-images.githubusercontent.com/6899256/219949227-ae40335b-dcbb-4392-9021-485f4f394925.mov

Also finding the top most viewcontroller was necessary to always be presentable. Removing that causes some links to not open. For example when a link is in a presented sheet.

That's why fixes like this are necessary to circumvent the In-App Browser preference. E.g on the About Fields in the Account Detail View (fixed as the about fields are no longer in a sheet)

```swift
UIApplication.shared.open(url)
return .handled
```

The Safari part was actually not laggy itself, but since it was presented in fullscreen via the top-most viewcontroller, it causes the underlying view to call `.onAppear` again where it then re-renders the view, causing the lags related to presenting it via the top-most viewcontroller.

This is the case for `IceCubesApp` where `onAppear` did setup RevenueCat and the push notifications again. Here it's simple enough to protect this with a `@State private var didAppear: Bool = false` as it is done on other views.

The other main issue is the `TimelineTab` where `onAppear` fetches the lists of the current account, causing it to re-render.

The current solution simply adds a `isSafariPresented` to the `RouterPath` to prevent loading the lists if we come back from Safari. But I think this could be removed if we solve the underlying issue, but for that I need some insight into the current logic of fetching the lists.

Is it necessary to fetch the lists every time the timeline tab is selected? Or could this be moved to one of the existing `onChange` modifiers, like account change or authentication change? Or maybe a new one?

Please let me know what you think, especially whether the current `isSafariPresented` solution is good enough or it should be optimized to work without it.

### DIFF
--- a/IceCubesApp.xcodeproj/project.pbxproj
+++ b/IceCubesApp.xcodeproj/project.pbxproj
@@ -919,7 +919,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 1.5.4;
+				MARKETING_VERSION = 1.5.5;
 				PRODUCT_BUNDLE_IDENTIFIER = "$(BUNDLE_ID_PREFIX).IceCubesApp.IceCubesNotifications";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
@@ -949,7 +949,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
-				MARKETING_VERSION = 1.5.4;
+				MARKETING_VERSION = 1.5.5;
 				PRODUCT_BUNDLE_IDENTIFIER = "$(BUNDLE_ID_PREFIX).IceCubesApp.IceCubesNotifications";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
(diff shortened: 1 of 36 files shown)


Dimillian:
I'm quite happy with the current behaviour. Not that about fields are not in a modal anymore I'm restoring the safarin handling in them (see latest commit).

Dimillian:
Not sure it's worth to re add all that + all the possible side effect just for the pretty transition. I agree it's 1500% better ... but this was causing such a lag and all the view to trigger their appearance. Your flag add a good amount of safety but also a good amount of potential issues :/

divadretlaw:
Yeah :/ the potential side effects are not great, since it can be easily forgotten in future changes to underlying views.

Another possibility would be to move the `SFSafariViewController` in its own `UIWindow`, thus preventing a `onAppear`/`onDisappear` on the underlying view. I'm gonna push a working version with the `SFSafariViewController` being in its own window. Not sure if that is worth it either, but why not 😅

Dimillian:
This is super smart!

PR #950, opened 2023-02-19
Optimize avatar view

kean:
The avatars in Mastodon seem to be standardized to 400x400px size, e.g. https://files.mastodon.social/accounts/avatars/108/194/113/520/602/290/original/a05c41565ddbe779.jpeg. The default display size in the feed is 40x40pt. With this change, the resizing is performed in the background and the bitmapped image exactly matches the display size.

The placeholder is currently called at least twice: one one the first body calculation and the second time on `onAppear` when the `LazyImage` starts loading. By extracting  `AvatarPlaceholderView`, its body is calculated only once – fewer reloads.

> **Warning**:  I smoke-tested the timeline and not much else


Dimillian:
Thanks! Tested around, seems all good.

PR #956, opened 2023-02-19
Trim whitespace and newlines in instance rules list

henrik:
It has bothered me that the first of the https://ruby.social rules has a trailing newline, which makes it render inconsistently.

Screenshot 2023-02-19 at 21 23 23

This should fix it – though I didn't set up and build the project locally for this tiny fix, so please do confirm 😅 

To be clear, this intentionally removes leading and trailing whitespace and newlines but not internal such: https://swiftfiddle.com/ykulagxc5bhsto52qqpcelztty


henrik:
Thanks! ❤️

PR #960, opened 2023-02-20
Font picker consistency

xurble:
This will move the option for SF Rounded from its own toggle into the popup menu with Hyperlegible and Dyslexia.

### DIFF
--- a/Packages/DesignSystem/Sources/DesignSystem/Font.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/Font.swift
@@ -15,11 +15,18 @@ public extension Font {
   private static let onMac = ProcessInfo.processInfo.isiOSAppOnMac
 
   private static func customFont(size: CGFloat, relativeTo textStyle: TextStyle) -> Font {
+    
+    
     if let chosenFont = UserPreferences.shared.chosenFont {
-      return .custom(chosenFont.fontName, size: size, relativeTo: textStyle)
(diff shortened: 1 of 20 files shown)


Dimillian:
You probably need to get main :p

Dimillian:
Raise a bug, get a commit lol

Packages/Env/Sources/Env/UserPreferences.swift
```diff
@@ -156,3 +153,16 @@ public class UserPreferences: ObservableObject {
     recentlyUsedLanguages = Array(copy.prefix(3))
   }
 }
+
+
+
+// https://stackoverflow.com/a/67305343
+extension UIFont {
```
Dimillian, line 160: Does it needs to be in two places?

Packages/DesignSystem/Sources/DesignSystem/Font.swift
```diff
@@ -66,3 +73,14 @@ public extension Font {
     customFont(size: userScaledFontSize(baseSize: caption), relativeTo: .caption)
   }
 }
+
+
+
+extension UIFont {
```
Dimillian, line 79: Here too

xurble: No. That's weird. Let me sort it along with the merge conflict.

xurble:
This should be in proper shape now.

Dimillian:
Indeed much better

PR #964, opened 2023-02-20
Fix building on Xcode 14.3 beta

Havhingstor:
Two optional bindings without optionals are now removed. Xcode 14.2 allowed this, but it's a compiler-error in the new beta. Additionally, Xcode complained about @MainActor being only allowed since iOS 13 in RevenueCat, so RevenueCat is updated to the current HEAD of main, where this problem doesn't persist.
I've also added a feature request to revenue-cat/purchases-ios to bump the version-number, so we should later be able to revert to using that version-number instead of the branch.


Dimillian:
I would hold off on that. I run the beta locally just to try it out, already made improvements on Sendable on main, but it'll probably be better to wait for some later betas before having those in main.

Havhingstor:
OK, no problem

PR #967, opened 2023-02-20
Fix display of multiple vote polls

### LINKED ISSUE #952: Bug: Multiple choice polls show incorrect numbers
## Environment:

 - OS: 
 
iPad OS 16.3.1
 - IceCubesApp version: 
 
1.5.2

## Description

If a poll allows multiple choice responses, Ice Cube displays the results incorrectly. Numbers are scaled so that results add up to 100%, which does not make sense for multiple choice polls. 

An example: https://androiddev.social/@cliffwade/109886759066153225

The Mastodon web interface displays results correctly. Percentages add up to more than 100% because people could choose multiple options. 

162F0432-88B7-4C9D-A6AA-31FA410437B2

Ice Cubes shows incorrect numbers that add up to 100%.

443054A3-A179-4BCE-A86C-FA659E206B6B

sothawo:
Closes #952

Besides the fix: In the small footer below a multiple choice poll this now display the number of voters along with the number of votes.

grafik


Dimillian:
Nice!

Issue #918, opened 2023-02-17
Crash on clicking to private profile


alper:
I just had a reliable crash browsing to a follower's private profile.

Latest iOS version of Icecubes.

Dimillian:
Do you have a post with the link to the profile or anything similar?

Dimillian:
Anyone can reproduce that? I've not been able to ever reproduce it.

alper:
I had it happen with another profile but going back wouldn't reproduce it. 

Do you have crash reporting installed? That should surface it, if it's more than a fluke.

Dimillian:
I do have it and for now I don't see anything related to profile. Or anything new by the way. This is weird as it's not the first report.

alper:
This one does it for me now a couple of times in a row: https://mstdn.social/@marick/

alper:
Now I have a video. First open no crash. Do some stuff click profile boom.

alper:
Here's the video: 
https://user-images.githubusercontent.com/72429/219871052-b5b03402-54e7-4a6a-b825-164d6f670e7a.mov

stodge:
I experience crashes 95% of the time when viewing the profile for @nakedcardboard@wargamers.social. It’s not a private profile but I thought I’d post I hear in case it helps identify a bug.

I’m on iOS 16.1.2.

alper:
Maybe that's right then. I got stuck on that version because the iPhone was full. Upgraded now.

Dimillian:
It's an iOS bug at this point. I've confirmed it. Related to iOS 16.1.X. Please update to 16.3 when you can. I'll soon close this one.

stodge:
Ooops, good to know. I have a brand new phone and I thought it had already updated to the latest iOS.

PR #972, opened 2023-02-21
Fix crash when opening account details

kean:
For me, the app consistently crashes on a device (iOS 16.1.1) when I open account details from notifications where "also followed by" section isn't empty. It crashes with the following issue:

Screenshot 2023-02-20 at 9 06 57 PM

I started reviewing the code and noticed `Group`. I had issue with it before, so I decided to remove it, and, sure enough, the crash was gone.  It looks like it's the combination of `Group` and these three `listRow` modifiers that leads to a crash:

```swift
        Group {
          makeHeaderView(proxy: proxy)
            .padding(.bottom, -20)
          familiarFollowers
          featuredTagsView
        }
          .listRowInsets(.init())
          .listRowSeparator(.hidden)
          .listRowBackground(theme.primaryBackgroundColor)
```

It doesn't crash on iOS 16.2 simulator, so presumably this issue with `Group` is fixed. But it's prone to issues and I personally avoid using it, especially directly inside `List` or other containers.

**Update:** I upgraded to iOS 16.3.1 and confirm that it no longer crashes.

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailView.swift
+++ b/Packages/Account/Sources/Account/AccountDetailView.swift
@@ -40,15 +40,13 @@ public struct AccountDetailView: View {
   public var body: some View {
     ScrollViewReader { proxy in
       List {
-        Group {
-          makeHeaderView(proxy: proxy)
-            .padding(.bottom, -20)
-          familiarFollowers
-          featuredTagsView
-        }
-        .listRowInsets(.init())
-        .listRowSeparator(.hidden)
-        .listRowBackground(theme.primaryBackgroundColor)
+        makeHeaderView(proxy: proxy)
+          .applyAccountDetailsRowStyle(theme: theme)
+          .padding(.bottom, -20)
+        familiarFollowers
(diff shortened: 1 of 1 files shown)


Packages/Account/Sources/Account/AccountDetailView.swift
```diff
@@ -487,6 +483,15 @@ public struct AccountDetailView: View {
   }
 }
 
+private extension View {
+  func applyAccountDetailsRowStyle(theme: Theme) -> some View {
```
kean, line 487: I'm open to suggestions how to implement/name it differently.

Dimillian:
So this is fixing this https://github.com/Dimillian/IceCubesApp/issues/918
I didn't had iOS 16.1 nearby, nice you found it!

Packages/Account/Sources/Account/AccountDetailView.swift
Dimillian: Name is fine!

PR #974, opened 2023-02-21
Fix status actions on boosting posts

te6-in:
- Fix the context menu showing edit, pin, and delete buttons on posts that I boosted but didn't post
- Fix the context menu ***not*** showing edit, pin, and delete buttons on posts that I posted and someone else boosted.
- Allow boosting my followers-only posts (which is available on Mastodon web)
  - Show `Boost to Followers` instead of `Boost` on context menus
  - Enable the inline boost button for my followers-only posts
  - But I'm not sure about the icon. Maybe there's something better than `lock.rotation`
- Disable the boost context menu from posts that can't be boosted (mentioned-only posts or others' followers-only posts)

I think the same mechanism should be applied to the swipe actions. But this is all for now.


Dimillian:
Nice

PR #977, opened 2023-02-21
Fix more button being too difficult to tap

te6-in:
Before:
SCR-20230221-lv5

After:
SCR-20230221-m6t


Dimillian:
I just did the same exact commit on main 😆

PR #978, opened 2023-02-21
Singularize enum type names

laclouis5:
Just a minor type name renaming suggestion for the sake of bike-shedding:

https://github.com/Dimillian/IceCubesApp/blob/b643dda0834148c5c01be70696691ebddcf3fb73/Packages/Env/Sources/Env/Router.swift#L6-L21

The Swift Programming Language [book](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/enumerations/) recommends using a **singular** name rather than a plural one:

> Each enumeration definition defines a new type. Like other types in Swift, their names (such as CompassPoint and Planet) start with a capital letter. Give enumeration types singular rather than plural names, so that they read as self-evident […]

The benefit of this naming convention can be seen in:

https://github.com/Dimillian/IceCubesApp/blob/b643dda0834148c5c01be70696691ebddcf3fb73/Packages/Env/Sources/Env/Router.swift#L72-L74

This function signature looks like it will navigate to multiple multiple destinations in a single call while it really does is to navigate to a single destination. The clarity benefit of a singular name appears in several other locations such as here where the reader can be confused into thinking that the `presentedSheet` variable will old multiple destinations:

https://github.com/Dimillian/IceCubesApp/blob/b643dda0834148c5c01be70696691ebddcf3fb73/Packages/Env/Sources/Env/Router.swift#L68

Using a singular name will also make call sites such as the following one more consistent by using a singular name both for the type name and the associated variable name `destination`: 

https://github.com/Dimillian/IceCubesApp/blob/b643dda0834148c5c01be70696691ebddcf3fb73/IceCubesApp/App/AppRouter.swift#L16

In this PR, I singularized most enum type names excluding `CodingKeys` enums and `Constants` enums which behave more like namespaces. I also excluded everything conforming to `Endpoint` since I wasn't sure about the semantic of these enums.

### DIFF
--- a/IceCubesApp/App/AppRouter.swift
+++ b/IceCubesApp/App/AppRouter.swift
@@ -13,7 +13,7 @@ import Timeline
 @MainActor
 extension View {
   func withAppRouter() -> some View {
-    navigationDestination(for: RouterDestinations.self) { destination in
+    navigationDestination(for: RouterDestination.self) { destination in
       switch destination {
       case let .accountDetail(id):
         AccountDetailView(accountId: id)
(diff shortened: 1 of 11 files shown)


Dimillian:
I never know, and I'm undecided when making an enum hehe!

PR #989, opened 2023-02-21
Fixed compact layout for video tip when video autoplay is disabled

Dean151:
Layout was broken by the too big sized Symbol.

Before - After:

before - after


Dimillian:
Thanks!

PR #991, opened 2023-02-21
fixes 965

### LINKED ISSUE #965: Bug: 2 bugs with messages with multiple photos
## Environment:

 - OS: 
 iPad OS 16.3.1
 - IceCubesApp version: 
 V1.5.4

## Description

I’ve found these 2 bugs when there is a message with multiple images:

- If you tap a photo, an incorrect photo is shown. Example: if you tap the first photo, the third one is shown.
- If you tap a photo, tap on the photo list (icon a the top left), choose another photo and go back to the message, the whole app becomes unresponsive.

https://user-images.githubusercontent.com/24291435/220177394-ee42e7e1-5038-4364-8468-b815a9f099fd.MOV

JulianKahnert:
Hi @Dimillian,
first of all: Thank you for this great app! It not only serves as an awesome SwiftUI example, but also is a great Mastodon client 🙏

I think these changes will fix #965 ... let me know what you think 🙂

### DIFF
--- a/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
+++ b/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
@@ -279,6 +279,8 @@ public struct StatusRowMediaPreviewView: View {
         .frame(maxWidth: isNotifications ? imageMaxHeight : nil)
         .frame(height: imageMaxHeight)
       }
+      // #965: do not create overlapping tappable areas, when multiple images are shown
+      .contentShape(Rectangle())
       .onTapGesture {
         Task {
           await quickLook.prepareFor(urls: attachments.compactMap { $0.url }, selectedURL: attachment.url!)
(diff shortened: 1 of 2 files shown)


Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
```diff
@@ -279,6 +279,8 @@ public struct StatusRowMediaPreviewView: View {
         .frame(maxWidth: isNotifications ? imageMaxHeight : nil)
         .frame(height: imageMaxHeight)
       }
+      // #965: do not create overlapping tappable areas, when multiple images are shown
+      .contentShape(Rectangle())
```
Dean151, line 283: Not sure how, but it sure fixes the issue!

Dean151 (approved):
LTM ; even tough I'm not sure if it exists a better solution, this one fixes it

Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
Dimillian: it indeed does make sense. I've never checked but the images were just leaking outside of their boundaries.

PR #996, opened 2023-02-22
Remove capture mode for images

kean:
The `isInCaptureMode` option appears to be unnecessary for avatars and images. `ImageRender` takes a screenshot after `onAppear` is called, so `LazyImage` is given a chance to fetch the image from the memory cache, which happens synchronously.

Prior to Nuke 12, `LazyImage` was using `UIImageView` for rendering. I suspect `isInCaptureMode` was added as a workaround for the following issue:

> ImageRenderer output only includes views that SwiftUI renders, such as text, images, shapes, and composite views of these types. It does not render views provided by native platform frameworks (AppKit and UIKit) such as web views, media players, and some controls. For these views, ImageRenderer displays a placeholder image, similar to the behavior of [drawingGroup(opaque:colorMode:)](doc://com.apple.documentation/documentation/swiftui/view/drawinggroup(opaque:colormode:)?language=swift).
> 
> From [`ImageRenderer`](https://developer.apple.com/documentation/swiftui/imagerenderer) docs.

Btw, starting with [Nuke 12.0 Beta 5](https://github.com/kean/Nuke/releases/tag/12.0.0-beta.5), `LazyImage` no longer even requires `onAppear` to be called to fetch the memcached images; it happens on the first `body` compute, so there is now even fewer `body` redraw cycles.

### How I tested it

- Select a post and tap "Share / Share post as image"
- Verify that the saved images has both the avatars and the content media (images)

### DIFF
--- a/Packages/Models/Sources/Models/Status.swift
+++ b/Packages/Models/Sources/Models/Status.swift
@@ -57,12 +57,12 @@ public protocol AnyStatus {
 
 public struct StatusViewId: Hashable {
   let id: String
-  let editedAt: ServerDate?
+  let editedAt: Date?
 }
 
 public extension AnyStatus {
   var viewId: StatusViewId {
-    StatusViewId(id: id, editedAt: editedAt)
+    StatusViewId(id: id, editedAt: editedAt?.asDate)
   }
 }
 
--- a/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
+++ b/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
@@ -171,7 +171,7 @@ public struct StatusRowMediaPreviewView: View {
               .frame(width: newSize.width, height: newSize.height)
           }
         }
-        .processors([.resize(size: .init(width: newSize.width, height: newSize.height))])
+        .processors([.resize(size: newSize)])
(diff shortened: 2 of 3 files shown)


Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
```diff
@@ -21,7 +21,6 @@ public struct StatusRowMediaPreviewView: View {
   public let isNotifications: Bool
 
   @State private var isQuickLookLoading: Bool = false
-  @State private var width: CGFloat = 0
```
kean, line 24: It was unused.

Packages/Models/Sources/Models/Status.swift
```diff
@@ -57,12 +57,12 @@ public protocol AnyStatus {
 
 public struct StatusViewId: Hashable {
   let id: String
-  let editedAt: ServerDate?
```
kean, line 60: I pushed this minor change together with the rest of the PR. I noticed `StatusViewId` was still showing up in the Profile, now because of ARC. I decided to try embedding `Date` directly into the `StatusViewId`, and it no longer does. I'm not sure how to explain it, but it's most like some Swift optimization.

Dimillian:
It's indeed work without it now! Nice!

PR #1000, opened 2023-02-22
Clean up settings capitalization

vtourraine:
Update EN `Localizable.strings` to follow title-style capitalization, per Apple guidelines (see the system Settings app for reference).


Dimillian:
Definitely a good move!

PR #1005, opened 2023-02-22
Add flair to status action buttons

scalbatty:
- makes tintColor viewModel independent in Action
- adds isOn function to Action
- moves actionButton to its own function for clarity (and help compilo)
- moves the counter outside the button
- creates StatusActionButtonStyle that defines how an action button behaves when tapped and toggled
- adds nested SparklesView that animates sparkles when the action button is tapped

Sidenote : couldn't get the "bouncy" scale effect I wanted. It wouldn't work on an iOS device, but did on the simulator.

https://user-images.githubusercontent.com/520924/220598379-c6f5f831-8edf-4b05-85b1-b229d5fd07c1.mov

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusActionButtonStyle.swift
+++ b/Packages/Status/Sources/Status/Row/StatusActionButtonStyle.swift
@@ -1,0 +1,128 @@
+import UIKit
+import SwiftUI
+import DesignSystem
+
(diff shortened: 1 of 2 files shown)


Dimillian:
This is wonderful 🌟 🌟 🌟 😻

Dimillian:
Working good and the code is awesome! Let's go!

PR #1008, opened 2023-02-22
Don't show the source-language before translating

Havhingstor:
I've removed the source language from the translate buttons to more accurately show what is done (The transition service automatically detects the language) and to not confuse the user if the poster sets the wrong source language.
This set language is now only used to determine if the button is shown in the main view in addition to the context menu. Furthermore, the detected source language is now visible in the box with the translated text and the translation service.


Dimillian:
Much cleaner!

PR #1012, opened 2023-02-22
Instance Info view refinements

icod:
- Format instance version stats in monospace to improve readability
- Format stats with thousand separators
  - Users
  - Posts
  - Domains

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/InstanceInfoView.swift
+++ b/IceCubesApp/App/Tabs/Settings/InstanceInfoView.swift
@@ -28,10 +28,12 @@ public struct InstanceInfoSection: View {
       LabeledContent("instance.info.name", value: instance.title)
       Text(instance.shortDescription)
       LabeledContent("instance.info.email", value: instance.email)
-      LabeledContent("instance.info.version", value: instance.version)
-      LabeledContent("instance.info.users", value: "\(instance.stats.userCount)")
-      LabeledContent("instance.info.posts", value: "\(instance.stats.statusCount)")
-      LabeledContent("instance.info.domains", value: "\(instance.stats.domainCount)")
+      LabeledContent("instance.info.version") {
+        Text(instance.version).monospaced()
+      }
+      LabeledContent("instance.info.users", value: format(instance.stats.userCount))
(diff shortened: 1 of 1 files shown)


sothawo:
This does not honour the local language setting,. When switching to German, I get this screen

image

In Germany we use the dot as a thousand-separator, the comm is for fractions.

IceCubesApp/App/Tabs/Settings/InstanceInfoView.swift
```diff
@@ -44,4 +46,10 @@ public struct InstanceInfoSection: View {
       .listRowBackground(theme.primaryBackgroundColor)
     }
   }
+  
+  private func format(_ int: Int) -> String {
+    let formatter = NumberFormatter()
+    formatter.numberStyle = .decimal
```
Dimillian, line 52: It's supposed to follow the current user locale no?

sothawo: Don't know if the NumberFormatter should do this. BUt setting the language in the simulator to German shows the formatting I captured in the screenshot, and that's definitely not german.

icod: I thought that this was assumed by the formatter but apparently not. 

I'll add the `formatter.locale = .current` later. I have seen this work for Dutch locale in the simulator, which, like German, uses . as thousands separator.

icod: Ah, wait. It's not the language, it's the region setting. Did you change that to Germany, @sothawo ?

Dimillian: And while we're here, would be good to have this formatter for the followers / following / posts count on user profile too now?

sothawo: Aaaaaaarghh. I changed the language not the region. Setting the region to Germany, it's alright. Sorry for the false alarm

Dimillian:
Ok merging it then!

PR #1013, opened 2023-02-22
Move autoPlayVideo setting to content settings view

icod:
The settings 'Auto-Play Video' belongs to the Media section of the Content Settings view, I think.

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
@@ -20,6 +20,9 @@ struct ContentSettingsView: View {
       }.listRowBackground(theme.primaryBackgroundColor)
 
       Section("settings.content.media") {
+        Toggle(isOn: $userPreferences.autoPlayVideo) {
+          Text("settings.other.autoplay-video")
+        }
         Toggle(isOn: $userPreferences.showAltTextForMedia) {
           Text("settings.content.media.show.alt")
         }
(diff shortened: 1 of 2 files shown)


IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
```diff
@@ -20,6 +20,9 @@ struct ContentSettingsView: View {
       }.listRowBackground(theme.primaryBackgroundColor)
 
       Section("settings.content.media") {
+        Toggle(isOn: $userPreferences.autoPlayVideo) {
+          Text("settings.other.autoplay-video")
```
sothawo, line 24: I think this should then change the localization key to `setting.content.media.autoplay-video`

Dimillian: Make it harder to keep track for when I'll merge adding other localization, like the ongoing Ukrainian one. I'm in favor of not moving the key if the value didn't change or/and still mean the same thing.

sothawo: Valid point. But the localizations need some clean-up at some time I think, I noticed, that in some languages they are grouped differently or have empty lines for grouping whereas in other languages this is not the case.

Dimillian: I do agree, and it would be good to have a common (free) tools to do that but I'm not aware of one :(

PR #1015, opened 2023-02-22
Always show buttons on focused status in details view

danieldickison:
I think it's helpful to have the buttons and their numeric badges visible when showing a status detail view for the focused status, even if the user has opted to hide them in timeline display settings.


Dimillian:
I agree!

PR #1037, opened 2023-02-23
collapse long posts

### LINKED ISSUE #914: Feature request: Collapse long post
I think a great feature, which I've seen in other apps like Ivory, Toot!, Trunks or Tusker, is the ability to collapse long posts, so that you have to press something like "Read More" to read the whole text.

sothawo:
Hi @Dimillian,

this is the first draft for the feature to show a collapsed version of long posts (I sometimes have posts in my timeline that are 4-5 screens long).

In order to determine if a post should be displayed collapsed, I use three values (they are in this draft PR hardcoded in the `StatusRowViewModel`)
1. a flag if the user wants to have posts collapsed at all
2. the number of characters a post must have to be collapsed
3. the number of lines to show for the collapsed post

I did not take into account the screen size or the device or try to guess the number of words to display. As this is done on the posts in the timeline, that calculation should be as fast as possible.

In the PR I have set pretty low values so that many posts are collapsed, this would later need higher default values and could be adjusted by the user.

The settings UI is still missing of course and so are the localizations. Before doing that I'd like to have your opinion about this solution.

And the are some UI things, like the collapse button not being as far left as I think it should be or animations when changing between the two states.

Running on my iPhone 13 mini I did not notice a slow down in scrolling.

Closes #914


Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
```diff
@@ -6,17 +6,40 @@ struct StatusRowTextView: View {
...
+  @ViewBuilder
+  func makeCollapseButton() -> some View {
+    if viewModel.showCollapseButton {
+      HStack {
+        Button {
+          withAnimation {
+            viewModel.isCollapsed.toggle()
+          }
+        } label: {
+          Text("Display " + (viewModel.isCollapsed ? "more" : "less"))
```
Dimillian, line 35: I think it could just be read more and maybe you don't even need to collapse it back?

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -37,6 +37,21 @@ public class StatusRowViewModel: ObservableObject {
   @Published var localStatusId: String?
   @Published var localStatus: Status?
 
+  // the next three values will come from user settings
+  let collapseLongPosts = true
+  let collapseThresholdLength : Int = 300
+  let _collapsedLines: Int = 5
```
Dimillian, line 43: I think we coud just have a configurable line limit with a default to 7 or 8 and that could be it. I don't think we need characters count. Lines limit scale well with the device size.

Dimillian:
I'll test tomorrow morning but I left two comments. I think we should go very simple with this one and just have lines limit and a read more label (like the new Twitter long post thing) at the bottom to expand it.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
sothawo: The character count is to determine if the collapse should be shown at all.  How would I know that a text exceeds the line count for the collapsed view and the button needs to be shown?

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
sothawo: "Read more" is better, yes. I notice when scrolling, that state in the viewmodel is reset once the post is scrolled out a bit and then returned into view, kind of autocollapsing then

EvilOne: I would suggest using **Show** more.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
Dimillian: 🤦🏻 Oh right. Yes I don't see a way to know if we're over the line limit or not without it costing too much performance.

sothawo:
Been thinking if we need this to be configurable at all. We could set the char limit for post to collapse to 700 or 800 and the number of lines to 7 or 8. Then display the "Show more" (or "Show full post"?) and skip the button to collapse again.

Other adaptions could be made when requests come in.

Instead of the theme tinted button, the expand label could look like the CW warning?

sothawo:
I set the default values in the model, and removed the collapse after expanding and adjusted the style of the expand link to match that of the content warning button.

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
```diff
@@ -6,17 +6,47 @@ struct StatusRowTextView: View {
...
+          withAnimation {
+            viewModel.isCollapsed.toggle()
+          }
+        } label: {
+          Image(systemName: "chevron.down")
+        }
+        .buttonStyle(.bordered)
+        .accessibility(label: Text("status.show-full-post"))
+        .accessibilityHidden(true)
+      }
```
EvilOne, line 44: Without this, one can tap between the words and the button on the far right and it will take them to the details page instead of expanding the post.
```suggestion
      }
      .contentShape(Rectangle())
```

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
```diff
@@ -6,17 +6,47 @@ struct StatusRowTextView: View {
...
+          })
+        Spacer()
+      }
+      makeCollapseButton()
+    }
+  }
+  
+  @ViewBuilder
+  func makeCollapseButton() -> some View {
+    if viewModel.showCollapseButton && viewModel.isCollapsed {
```
EvilOne, line 28: Could we move this entire logic to viewmodel?

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
sothawo: thx, will adjust that - and fix it in `StatusRowSpoilerView` as well where I copied my code from ;-)

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
sothawo: yep, and cleaned up the property getters a little

Dimillian:
I've pushed one minor fix but I think I still need to think a bit about it, a few notes: 

* I think we definitely need a content setting to turn it off. I always want post auto expanded for example.
* Maybe in the detail view all posts should be auto expanded? I don't know, I feel like it should. 
* If there is a content warning before, you need to first tap the expand content warning then this new expand post. Again, is this what we want.

sothawo:
> I've pushed one minor fix but I think I still need to think a bit about it, a few notes:
> 
>     * I think we definitely need a content setting to turn it off. I always want post auto expanded for example.

Ok, that's a boolean in the content setting with perhaps an explanatory text, I'll add that

> 
>     * Maybe in the detail view all posts should be auto expanded? I don't know, I feel like it should.

That's the hardest for me to implement (still new to SwiftUI) - figuring out in which context the view currently is displayed.  But a valid point, the web ui when showing the full text of a post that's marked with "Read more" switches to the detail view to do it - not that I want to switch to detail, but as you said, in detail view the post should be expanded already.

> 
>     * If there is a content warning before, you need to first tap the expand content warning then this new expand post. Again, is this what we want.

hm, I for example have CW content autoexpanded but would nevertheless like the collapse feature. But if I then see a post that had a cw and is expanded and I don't want to see it, I can just use the cw button to collapse it.

Dimillian:
* Yes nice. There is also the need to connect it properly, it need to be done at the init time of the VM but we barely have access to preferences there. Let me know if you have troubles.
* Let me do it, need to careful too because we don't want to create unnecessary update cycle in the timeline or in the detail view.
* I don't think this is a big issue anyway, we can re visit later.

Dimillian:
Sound fun :p Take your time, I'll be a bit away this evening and this weekend, for once :)

sothawo:
addressed the first one. There now is a setting on the content tab. Reworked the model and how the view passes this setting into the model. Works for me now (in the suimulator)

sothawo:
third point done. Posts with a CW are not collapsed by this logic

Packages/Env/Sources/Env/UserPreferences.swift
```diff
@@ -47,6 +47,8 @@ public class UserPreferences: ObservableObject {
   @AppStorage("swipeactions-icon-style") public var swipeActionsIconStyle: SwipeActionsIconStyle = .iconWithText
 
   @AppStorage("requested_review") public var requestedReview = false
+  
+  @AppStorage("collapse-long-posts") public var collapseLongPosts = true;
```
Dimillian, line 51: ; really? hahaha long time no see the semicolon

sothawo: shame on me. (Spring Data Elasticsearch is Java, in my business project we're using Kotlin). BUt you have them in the localization files as well!

sothawo: fixed

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
```diff
@@ -1,22 +1,61 @@
...
+          .emojiSize(Font.scaledBodyPointSize)
+          .environment(\.openURL, OpenURLAction { url in
+            viewModel.routerPath.handleStatus(status: status, url: url)
+          })
+        Spacer()
+      }
+      makeCollapseButton()
+    }
+    .onAppear {
+      viewModel.collapseLongPosts = preferences.collapseLongPosts
```
Dimillian, line 27: I'll do a few modifications here. Basically this cost a lot of performances. onAppear modifying a published property (even if it's set to the same value) trigger a body calculation of everything observing the viewModel. Will probably set it only if the preference is different from the VM value.

sothawo: Is there a different way to set the initial value into the viewModel? I was not sure about acccessing the `UserPreferences` from the model although in `markSeen()` this is done as well.

Dimillian:
I won't have time to rework it this morning or probably today. I think adding the preferences was necessary but made a lot of things more complexe. We can't really set the lineLimit everytime a status row appear. This trigger too many refresh and make the timeline slow to scroll yet again. Having it calculated in the init of the view model would be better. It's ok if it does not refresh live with a preferences change but only for the new posts.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -37,6 +37,36 @@ public class StatusRowViewModel: ObservableObject {
...
+      _collapseLongPosts = newValue
+      recalcCollapse()
+    }
+  }
+  
+  private func recalcCollapse() {
+    let hasContentWarning = !status.spoilerText.asRawText.isEmpty
+    let showCollapseButton = collapseLongPosts && isCollapsed && !hasContentWarning
+      && (status.reblog?.content ?? status.content).asRawText.unicodeScalars.count > collapseThresholdLength
+    lineLimit = showCollapseButton && isCollapsed ? collapsedLines : nil
```
Dimillian, line 67: This is basically the problem. It's called every time the row appears, triggering an update of a published property as the view appear.

Packages/Status/Sources/Status/Row/Subviews/StatusRowTextView.swift
Dimillian: You can access it using .shared as static. It's fine if we don't observe it.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
sothawo: is the update triggered even when the published property does not change? Then only setting the property if it really changed should make it better?

Dimillian: Yep. You can see the setter in the time profiler. Even if it doens't change anything the body shadow will be recalculated unfortunately.

sothawo: I removed the onAppear an onChange from the view, the user pref is read once in the init() of the model, and the published lineLimit is only set when it changed from the previous value

Dimillian:
This is good! A bit more static but much safer & simpler. Will merge 🚀

Issue #1044, opened 2023-02-24
Feature request: favoriting direct messages in dm tab


blueturtleai:
It should be possible to favorite a direct message directly in the direct message tab. Currently this can only be done in the notifications tab or the home timeline.

paul4565:
You can do this if you long press on the message inside the conversation. I had requested this as well.

blueturtleai:
Okay, thanks for the information! I‘d suggest, that favoriting is added to the … menu too.

blueturtleai:
Fixed in #1053

PR #1053, opened 2023-02-25
Enhance the context menu for private messages

Havhingstor:
The context menu of private messages in the chat view now allows bookmarking of messages, the creation of a mention of the poster of the message, and reporting of a message.


Havhingstor:
The user can now also interact with the last message of a conversation in the list with all conversations.
See #1044

The change to "Conversation" was necessary since just reloading the conversations would fetch the old list if done directly after sending the like / bookmark request (at least with my instance, I had to wait 0.5 seconds, which would probably be different for different instances), so we directly have to modify the affected conversation.

Dimillian:
Great set of features for the messages list!
I'll look into it a bit later but there is many problems with the way this is done. 
I don't think managing a list of child view models is the right way to do it, it's a first in the codebase and I would rather have a self managed top level list of conversations. We don't really need to have a VM per row. Or maybe then it could be built right in the view with State etc...
There is also a crash if you scroll down to get the next page and then go in and out of a conversation.
I might close this and rework it later, or maybe you can rework it if you have some idea to avoid the `ConversationsListRowModel` list and published property.

Havhingstor:
I'll try to rework it... Even if that doesn't work, I'd suggest keeping in the changes in the context menu for single messages, since that is quite similar to the already existing menu options.

Havhingstor:
I've removed the view model for the row, the conversation used is now a binding to an element of the list in the list view model.

Dimillian:
Better thanks!

PR #1065, opened 2023-02-25
Change icon for bot profile

sothawo:
Had a short conversation on mastodon where a user asked what the gear icon next to the profile name means. One of the comments came up with this one as it looks more like a robot (I never would have thought of using a poweroutlet symbol)

image

I like it better as well, your decision to use it ;-)

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailHeaderView.swift
+++ b/Packages/Account/Sources/Account/AccountDetailHeaderView.swift
@@ -137,7 +137,7 @@ struct AccountDetailHeaderView: View {
               .font(.scaledHeadline)
               .emojiSize(Font.scaledHeadlinePointSize)
             if account.bot {
-              Text(Image(systemName: "gearshape.fill"))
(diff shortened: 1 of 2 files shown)


Dimillian:
Ahahhahahahaha this is awesome! Don't forget to change it on the profile too

Dimillian:
I agree it's much better.

Dimillian:
I think you should use poweroutlet.type.f it's even better! And not sure it need to be filled?

Dimillian:
image
Crazy how all of them fit the bill

sothawo:
'f' even has ears!

thats 'f'

image

and that's 'f.fill'

image

looking at these I'd prefer to stick to 'b' in that small size the round with ears gets lost a bit

Dimillian:
Alright let's go!

sothawo:
it didn't push the 'f', so the PR is still on 'b'

Dimillian:
Don't forget this one :)

sothawo:
Oops, totally missed your first comment about the profile view. Adjusted that.

PR #1072, opened 2023-02-26
Resolve all possible Sendability warnings from Swift 6 compatibility mode

AlanQuatermain:
I turned on all the sendability warnings for Swift 6 mode and fixed every warning possible. There are a few which are un-fixable from here, which I'll outline below. There are several pieces here, dotted all over the place.

First of all, pretty much everything in Models is `Sendable` now. If I've missed anything, it's because it wasn't being passed across task boundaries anywhere. This was straightforward in just about every case, since just about everything is a value type; there were three exceptions, though, which necessitated special handling.

- `ReblogStatus` is a class. Happily, every property in there was immutable with the exception of `application`. That property is never written to anywhere, so I changed that from `var` to `let` and marked the whole thing `Sendable`.
- `Status` is also a class. Its contents are all immutable, with the exception of `userMentioned`, which is modified here & there. In this case, since it's a scalar type with atomic instruction support, I imported the Swift `Atomics` package and used an `UnsafeAtomic<Bool>` to hold the value in a private `let` property, with `userMentioned` acting as a wrapper. Since `UnsafeAtomic` isn't `Codable`, though, I manually copied in the compiler-generated implementations from before the change, then tweaked the decoder implementation to initialize the atomic value.
- `Client` was causing some problems, and there were several mutable variables here. I played with making it an `actor`, since most of its API is `async` anyway, but there were problems with protocol conformances; `id` had to be `nonisolated`, which means the same applied to `isAuth`, which was reading from a mutable property, etc. I wound up creating a structure to contain the mutable state and I stuck that into an `OSAllocatedUnfairLock`. All accesses to those variables go through that lock now. An added benefit is that `OSAllocatedUnfairLock` is `Sendable` when its state is also `Sendable`, so that left us with a provable conformance for `Client`, rather than having to use `@unchecked Sendable`. Nice.

Another thing that popped up was a number of complaints about `ObservableObject` and `Published` from SwiftUI being aliased to the versions in `Combine`, and that in Swift 6 it would be an error to use these without importing `Combine` manually. So there's now a whole bunch of `import Combine` floating around to fix those.

Some other places where sendability errors cropped up were in nested enum types within types tagged with `@MainActor`. The `allCases` implementation generated by the compiler doesn't include any isolation attributes, so it inherits `@MainActor` by default. But as noted above, protocol conformances can't be provided by actor-isolated implementations, so that was causing errors. The solution there was to manually implement `allCases` for each enum that had this problem. There are comments describing the issue attached to each instance, for posterity.

There are a few warnings remaining that I can't do anything with right now, and which crop up when calling certain async methods from within an actor (or code tagged `@MainActor`).  These are due to `async` methods from Foundation (and one from Nuke) that take protocol types as delegates, such as `(any URLSessionTaskDelegate)?`. The protocol in that case comes from Objective-C, so it's required to be implemented by a class, which means the compiler can't assume immutability, etc. To add insult to injury, these are usually in optional parameters that aren't used anywhere, so it looks as though the warnings are talking bollocks. I've filed radars about these cases, and I'm hopeful it'll no longer be an error in the next Xcode release, at least for invocations using the `nil` default values.

A similar issue occurs when using the async APIs on `NSItemProvider`. These take/return plain `NSDictionary` types, which aren't thread-safe, and thus are not sendable. When imported into Swift, these usually become `[AnyHashable : Any]`, which is a firehose of non-sendability. There are Radars out for that too, but it's a tough one—we don't want to limit all clients of `NSItemProvider` to transferring only Sendable types, but clearly there are cross-actor boundary issues to consider. Thus I wouldn't expect a simple change of attributes on this one.


AlanQuatermain:
I have no idea why it says my commit is 'unverified' and that no user is associated with the committer email. It's literally the main email on my profile page, and I've signed with the same GPG key I use for everything else on here. Sigh.

Dimillian:
This is huge! Really too bad we would have to go with a custom decoder for the status object. It's almost a deal breaker :/ Do you think it would be possible to be avoided? 

Also, I know that the Swift 6 Concurrency API is yet to be completely finalised. Do you think it's safe to make those changes now? I do think yes, but who knows if Apple backtracks on some of the design.

Packages/Models/Sources/Models/Status.swift
```diff
@@ -71,7 +72,11 @@ protocol StatusUI {
 }
 
 public final class Status: AnyStatus, Codable, Identifiable, Equatable, Hashable, StatusUI {
-  public var userMentioned: Bool?
+  private let _userMentioned: UnsafeAtomic<Bool>
```
Dimillian, line 75: I'm wondering if this could leave somewhere else. It could probably be a purely view and calculated in the init of the row view model. So we don't have to male the sacrifice of making this a custom decoder for this class.

Dimillian: Maybe it could be bested in a sub structure? So we would decode only this prop manually?

Packages/Network/Sources/Network/Client.swift
```diff
@@ -28,42 +36,52 @@ public final class Client: ObservableObject, Equatable, Identifiable, Hashable {
...
   private let decoder = JSONDecoder()
 
-  /// Only used as a transitionary app while in the oauth flow.
-  private var oauthApp: InstanceApp?
-
-  private var oauthToken: OauthToken?
+  // Putting all mutable state inside an `OSAllocatedUnfairLock` makes `Client`
+  // provably `Sendable`. The lock is a struct, but it uses a `ManagedBuffer`
+  // reference type to hold its associated state.
+  private let critical: OSAllocatedUnfairLock<Critical>
```
Dimillian, line 45: Ah this is nice. I was wondering how to make client mutable properties isolated. Do you think it's a good way to do this moving forward? Or should `Client` have a nested actor `ClientValues` or something like that?

Dimillian:
@AlanQuatermain I've moved `userMentioned` outside of `Status`, so you can probably clean the custom decoder and the dependency if it's not needed anymore.

AlanQuatermain:
All done—good call on moving `userMentioned` out of the model. I'll push another commit in a moment that removes the `Atomics` dependency.

AlanQuatermain:
> Also, I know that the Swift 6 Concurrency API is yet to be completely finalised. Do you think it's safe to make those changes now? I do think yes, but who knows if Apple backtracks on some of the design.

It's safe to make these changes here. The type of things the new warnings are pointing out are all based around data races, which is a key part of Swift 6. If anything there will be sterner errors cropping up, or more tightly-targeted ones.

Packages/Network/Sources/Network/Client.swift
AlanQuatermain: A nested `actor` type would work, though you'd have to do some dancing around it to make things like `id` remain synchronous. In this case I'd favour the lock, as it will perform the same job with a little less overhead and a lot less changes in the code & API.

Dimillian:
:ahhh: Sorry i've merged too early but made the commit remove Atomics just after. I thought it was already pushed 😆

AlanQuatermain:
No worries—I've pulled the latest changes over to my side of the wall & it all looks good.

I'm going to have a run through Instruments next and see if I can't pin down some performance enhancements around the status timeline; chiefly things like `StatusRowView`, which holds references to a lot of big `ObservableObject` properties, and has a large number of pieces that might be best broken out into subviews.

One useful thing to know that isn't necessarily obvious is that it's cheaper to take chunks of `body` code and put them into entirely separate `View` instances rather than moving the code into dynamic properties or functions. When that code is in a function, it's guaranteed to run every time your view's `body` is run, but when it's in a subview, then only the initializer for that subview type is going to run—the SwiftUI runtime can determine whether it needs to evaluate the `body` already.

Secondly, it's worth remembering that, from the viewpoint of updating the dependency graph, `ObservableObject` types are essentially atoms. If *anything* changes in there, *every* `@Published` member is marked as updated; it's the nature of the beast. What I'd like to look into is whether this is actually causing any noticeable performance issues.

Dimillian:
Yep. I've already did a big cleanup on main in this row maybe you didn't had a chance to look at it yet. The view are split into a lot of small components now with very targeted updates. Basically as you scroll the timeline no update happen at all. The only things to save now I think, is most probably faster view body computation.

AlanQuatermain:
Oh nice! Yeah that looks like what I was planning to try out.

PR #1080, opened 2023-02-26
Add 2 new icons

te6-in:
This adds 2 new icons. They're almost the same as the old ones, but they're without the white borders.
If you don't like the size of the app becoming too big because of icons, just feel free to close this 🙂

스크린샷 2023-02-26 22 23 51

스크린샷 2023-02-26 22 23 26


Dimillian:
Awesome. Those have been requested a lot!

PR #1099, opened 2023-02-27
Change arrows icon to rocket for boosts

xurble:
Added some extensions in SFSymbols.swift.  They work by convention.  All lower case image names are assumed to be system SF Symbols.  Mixed case image names are assumed to be in the app's Asssets.


sothawo:
my personal opinion: I don't like this, it looks like mimicking ivory.

Dimillian:
It seems like there is few conflicts to fix. 
@sothawo I think it's still better than what we have right now. If I could I would put the 🚀 emoji haha.

sothawo:
a rocket is a better match for the word "boost" I have to admit

xurble:
Yeah the idea wasn't to ape Ivory so much as match the Boost notification emoji. 

Seemed better than replacing that
Rocket with 🔄

xurble:
I do think Mastodon made a mistake changing the terminology from retweet to boost and then not changing the iconography. Those arrows make sense for retweet or repost but not boost.

Packages/DesignSystem/Sources/DesignSystem/SFSymbols.swift
```diff
@@ -0,0 +1,34 @@
...
+import SwiftUI
+
+// Functions to cope with extending SF symbols
+// images named in lower case are Apple's symbols
+// images inamed in CamelCase are custom
+
+extension Image {
+  
+  public init(imageNamed: String) {
+    if imageNamed.lowercased() == imageNamed {
```
Dimillian, line 13: I wish we could avoid that or do it differently. This could be a bottle neck to to this str compare for each icons of the status row actions bar. It might seems nothing but when it's supposed to scroll at 120 FPS this could have an impact :/

Dimillian: All good now

PR #1113, opened 2023-02-28
Optimize images size

laclouis5:
The app currently contains many (36) PNG images (1024x1024) which are used for the app icon variants. These images are not directly used for the app icons but only to be presented in the user interface in the `IconSelectorView` (the app icons are stored separately in AppIcon assets catalogs).

Contrary to the AppIcon assets, theses images are not required to be in PNG format (not entirely sure of this). According to [Emerge Tools](https://docs.emergetools.com/docs/optimize-images#ios) (a great source for app size analysis), the images can be converted to HEIC to efficiently reduce their size without reducing the image quality significantly.

In this PR I converted the 36 images to HEIC (macOS default quality settings) which resulted in a 10 Mo decrease of the Assets catalog size. I don't have a compatible iOS app to test the app size but the gain should be similar.

Further size optimizations in this area may be possible (~ another 10 Mo):

* also convert AppIcon assets to HEIC if not required to be PNGs,
* entirely avoid duplicate images as currently there is one for the icon and one for the UI ([relevant](https://www.avanderlee.com/swift/alternate-app-icon-configuration-in-xcode/#changing-the-app-icon-in-swift)),
* decrease the resolution of the images (UI ones) if not necessary to be 1024x1024.


Dimillian:
Thanks for that!

PR #1114, opened 2023-02-28
Reducing the app size by stripping symbols

laclouis5:
Since Xcode 14, the app symbols are not automatically stripped anymore, which causes an app size increase. 

In this PR I implemented the suggestion made by [Emerge Tools](https://docs.emergetools.com/docs/strip-binary-symbols) by adding a build script which strips symbols.

I'm not entirely sure if that works since (1) this would invalidate the code signature of the app and (2) I did not checked app size gains on the real app. Thus, this is more a proof of concept to present the idea if more experienced devs want to investigate this path.

### DIFF
--- a/IceCubesApp.xcodeproj/project.pbxproj
+++ b/IceCubesApp.xcodeproj/project.pbxproj
@@ -771,6 +771,7 @@
 /* Begin PBXShellScriptBuildPhase section */
 		B906F02829AF6FCE001BCDD3 /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;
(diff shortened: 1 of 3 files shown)


Dimillian:
Hmmm I'm not really sure I would want that.

laclouis5:
If I understand well, it's just that you *may* have to re-codesign the binary after stripping, but I don't have enough experience with code-signing to be sure of that.

Dimillian:
Thanks! I'll close it for now because I'm not super fan of adding a custom build script that could add unforeseen issues in the future.

Issue #1069, opened 2023-02-25
Bug: Crash on macOS when attempting to add account


thattridentdude:
## Environment:

 - OS: 
 macOS 13.3
 - IceCubesApp version: 
1.5.9 / current main branch (2ba2675ae443c7a61e091684163d39c9d13160fe) 

## Description

When starting the app for the first time on macOS you're offered to add an account. 
Attempting to do this by trying to add a mastodon.online account crashes the app the moment it tries to hand off login to Safari. 

I'm currently still investigating and will provide PR if I can find cause and fix, but I've found out the following so far:

That said, this might be a separate issue because manually typing `mastodon.online` in full doesn't prevent the crash. 

Attached is a zip with console output (incl. stacktrace) and unsymbolicated crash reports from App Store version. 

## Related Issues

 - [X] Search that this bugs don't already exist before creating it.
 
[IceCubesCrashData.zip](https://github.com/Dimillian/IceCubesApp/files/10832373/IceCubesCrashData.zip)

-----

Edit: 

Removed console lines about incomplete url, is unrelated to this issue. 

Issue is caused by `SFSafariViewController` being broken on the macOS 13.3 dev beta, including the most recent one (22E5230e). 

13.3 [release notes](https://developer.apple.com/documentation/macos-release-notes/macos-13_3-release-notes) make no mention of any changes that could affect this, so this might be a bug. 

Nevertheless, I've created a workaround in #1117 and will file a FB with Apple. 

If this is new (intended) behavior or workaround is necessary long term for whatever reason, then I suggest polishing it up the workaround a bit.

Will reply with comment once/if I have more information about FB report to Apple.

thattridentdude:
Reported to Apple under FB12021800. 

Tried to create minimal reproducible example (see attached) but I'm unable to reproduce the issue in the example. 

I think I'm going to need help from someone who is more familiar with the codebase to see if we can add in some more elements from Ice Cube App and see if that might reproduce the issue. 


[macOS 13.3 SFSafariViewController Bug.zip](https://github.com/Dimillian/IceCubesApp/files/10856155/macOS.13.3.SFSafariViewController.Bug.zip)

Dimillian:
So PhotoPicker is broken in the same sort of way on iOS 16.4. Really hope it's just a in flight beta issue and will not ship like that. Will probably wait for RC before I upgrade myself. If at this point it's still like this, then we'll go with your fix.

thattridentdude:
[...]

PR #1117, opened 2023-03-01
Workaround for macOS 13.3 SFSafariViewController issue (Fix for #1069)

thattridentdude:
macOS 13.3 breaks SFSafariViewController, leading to a crash as described in [#1069](https://github.com/Dimillian/IceCubesApp/issues/1069).

It's unclear if this is an intended change by Apple, release notes make no mention of anything that could cause this issue and today's 13.3 developer beta (22E5230e) hasn't resolved the issue. Public release 13.2.1 doesn't have this issue. 

For now I've created a workaround in this commit and will submit a feedback report with Apple. Issue  [#1069](https://github.com/Dimillian/IceCubesApp/issues/1069) will be updated with more information on this feedback report.

Workaround could use some polish if this is going to be a long term solution.

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/AddAccountsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/AddAccountsView.swift
@@ -12,6 +12,7 @@ import SwiftUI
 struct AddAccountView: View {
   @Environment(\.dismiss) private var dismiss
   @Environment(\.scenePhase) private var scenePhase
+  @Environment(\.openURL) var openURL
 
   @EnvironmentObject private var appAccountsManager: AppAccountsManager
   @EnvironmentObject private var currentAccount: CurrentAccount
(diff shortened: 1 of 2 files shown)


Dimillian:
It's so weird, because it's supposed to open Safari on macOS, or actually any default browser you have and not a SFSafariViewController. I might want to wait a bit before merging that. It also missing the localization added in every other file. But I would not bother with it for now, just add the string inline in the code.

Dimillian:
Did you had a chance to try on the latest beta? Still relevant?

thattridentdude:
> Did you had a chance to try on the latest beta? Still relevant?

Somehow never got a notification of your comment, sorry about that. 

Doesn't seem to be relevant anymore. MAS version even works on Sonoma beta, so I think this can be closed.

PR #1119, opened 2023-03-01
Record quote toots that have failed to load so that we don't try and load them again.

### LINKED ISSUE #1118: Bug:  Some links incorrectly assumed to be quote toots
## Environment:

 - OS: 
iOS, macOS
 - IceCubesApp version: 
 1.5.12

## Description

If Ice Cubes interprets something as a quote toot that actually isn't one (such as if you follow non-mastodon fediverse accounts like micro.blog and they link to other posts on their server), that quote toot will never enter the cache and so every time you see it, it will be placeholdered, reloaded and removed.

Also when you try and click on the link to that post, the internal post handler will hijack the link and then fail.

https://user-images.githubusercontent.com/246116/222023455-fd166118-ae76-4fae-866c-dc41811d6ef8.mp4

## Related Issues

 - [ ] Search that this bugs don't already exist before creating it.

xurble:
Fixes #1118

1: Stops repeated visible insertion and removal of placeholder quote toot. 
2: Stops link hijacking of inline status viewer allowing links to be followed as regular URLs

### DIFF
--- a/Packages/Env/Sources/Env/Router.swift
+++ b/Packages/Env/Sources/Env/Router.swift
@@ -95,12 +95,14 @@ public class RouterPath: ObservableObject {
               client.hasConnection(with: url),
               let id = Int(url.lastPathComponent)
     {
-      if url.absoluteString.contains(client.server) {
-        navigate(to: .statusDetail(id: String(id)))
-      } else {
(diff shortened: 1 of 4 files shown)


Dimillian:
Awesome, just made a slight change to move it to the embed cache

PR #1130, opened 2023-03-01
Neon icons

xabirequejo:
I tried to create a neon resembling icon based on the original Ice Cubes app icon. I did it for practising mainly but since they are done I'm uploading them here. Feel free to delete.


Dimillian:
I like it! But not sure if I want to add yet one other icon yet :) will keep it around for later! Thanks!

Issue #1120, opened 2023-03-01
Bug: Pasting full instance URL incl. "https://"  in AddAccountsView prevents you from signing in with instance


thattridentdude:
## Environment:

 - OS: 
iOS/iPadOS 16.3.1 & macOS 13.2.1
 - IceCubesApp version: 
App Store 1.5.12 and current commit (57197b9e9a55359eead1cf4a189526f3a0804d6f)

## Description

When pasting the full URL of instance in AddAccountsView (e.g. "https://iosdev.space/"), it will pull up the instance info and offer you the "Sign in" button, but the button doesn't do anything. 

Console log seems to imply that something goes wrong with making the oath URL, but I'm not entirely sure as I'm still unfamiliar with the code base.
It might be as simple as copy pasting [AddAccountView.swift:85 and 86](https://github.com/Dimillian/IceCubesApp/blob/57197b9e9a55359eead1cf4a189526f3a0804d6f/IceCubesApp/App/Tabs/Settings/AddAccountsView.swift#L86) and pasting it somewhere where the oath URL is made.

Suggestion is to obviously fix the making of the oath URL, but also to graciously fail if something like this happens in the future.
Currently there's no indication towards user that something is wrong other than the button not doing anything form the user's POV. 

## Related Issues

 - [X] Search that this bugs don't already exist before creating it.

Dimillian:
I think we have a duplicate, but yes this is a bug.

thattridentdude:
I can't find the dupe, but feel free to close this one if you've found it.

PR #1131, opened 2023-03-01
Add Account: Sanitize the servername to remove http[s]:// and also usernames

xurble:
Fixes: #1120


Dimillian:
Hmmm it break down when you tap on a suggestion and replace the domain.

xurble:
Sorry, I thought instance.title was always the domain, but it's just mostly the domain.  Try now.

Dimillian:
Thanks!

thattridentdude:
Nice work!

PR #1141, opened 2023-03-02
Display and edit filter expirations

### LINKED ISSUE #1124: Feature request: Add optional expiry date to filters
Filters can have an optional expiry date. Currently icecubes does not display this or allow editing/setting.

@Dimillian I'm working on this.

sothawo:
## Filter durations

Expiration values that are set on filters now are displayed in the filter list:

image

When editing a filter, a new dropdown offers the different values:

image

When anything else than `infinite` is selected, a date picker is shown that reflects the selected entry (this is read only, except for the `custom` selection:

image

Custom selection allows setting a value:

image

I know that we normally don't hide controls that are inactive and only disable them, but with the datepicker this is a  problem: It's not possible to display a date picker with no date set which would be needed for the `infinite` entry. So for `infinite` we would show some random date which does not reflect the setting. So I chose to hide the datepicker for `infinite` setting.

## Consolidation of duration enums

All places that show some duration values (poll creation, account muting, filter expiration) now use the same `Duration` enum. This consolidates the numeric values for seconds to expire and the localizations in one place instead of duplicating them.

Closes #1124

### DIFF
--- a/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
@@ -25,10 +25,12 @@
 "enum.avatar-shape.circle" = "Кола";
 "enum.avatar-shape.rounded" = "Акруглены";
 "enum.durations.infinite" = "бясконцасць";
+"enum.durations.custom" = "custom";
 "enum.durations.fiveMinutes" = "5 хвілін";
 "enum.durations.thirtyMinutes" = "30 хвілін";
 "enum.durations.oneHour" = "1 гадзіна";
 "enum.durations.sixHours" = "6 гадзін";
+"enum.durations.twelveHours" = "12 гадзін";
 "enum.durations.oneDay" = "1 дзень";
 "enum.durations.threeDays" = "3 дні";
 "enum.durations.sevenDays" = "7 дзён";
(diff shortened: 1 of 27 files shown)


EvilOne:
UI looks good, few comments/suggestions.
- instead of infinite I would suggest Forever
- drop the custom option for now and see if people mention it. 

Thoughts?

EvilOne:
The infinite option doesn't need a date picker does it? Couldn't the code just handle that behind the scenes and set it to 100 years from now or something? And if the duration doesn't match any of our predetermined selections we know to display infinite/forever.

Does this make sense?

Dimillian:
Looking good! I guess you forgot a few spot to add the localization keys?
image

sothawo:
> Looking good! I guess you forgot a few spot to add the localization keys? 

Oops. Had them in for sure, but then when starting to add it to the others, I probably cut instead of copied them. Fixed. And I use en-GB on my phone and simulator setup so I didn't notice.

sothawo:
> UI looks good, few comments/suggestions.
> 
>     * instead of infinite I would suggest Forever

I don't mind. It's a word. The web ui has named this "Expire after"/"Never". BUt we have to make sure the localization fits all  places where it's used. That's filter and the mute account popup. So forever would be fine for me

>     * drop the custom option for now and see if people mention it.

I'll be the first to ask for it. Basically that's the main reason I just implemented it here, because that's one thing I miss in the Web UI.
My last use case: The football world championshiop. I don't care about football and even less about that event. So I ewanted to set up a filter that ends a day after the finals. I could'nt do that in the web ui, I had to fall back to using postman to set the filter via the API (manually calculating the expiresIn value).

Second use case: When a user set a value of 7 days and then one day later adds a new keywaord, what will be the expiratin value then? 5 days would be less than the original value, 7 be more. So in that case we'd need an option "don't change the existing value".

sothawo:
> The infinite option doesn't need a date picker does it? 

That's why I don't display one in that case.

> Couldn't the code just handle that behind the scenes and set it to 100 years from now or something? And if the duration doesn't match any of our predetermined selections we know to display infinite/forever.

A date in the future isn't infinite. I had tried that and it looks pretty strange to have some random date in the future displayed. And we have to consider the value to send to the server. Infinite means to set  no or clear out an already set value on the server.

I have tried these options in code,  and in my opinion they are worse than displaying or hiding the picker and using the custom entry for an already set value

Dimillian:
Well I like the custom with the native calendar! I think it's totally fine as of right now.

PR #1152, opened 2023-03-03
Updated french translation

enetter:


IceCubesActionExtension/ActionRequestHandler.swift
```diff
@@ -15,7 +15,7 @@ import Network
 // Sample code was sending this from a thread to another, let asume @Sendable for this
 extension NSExtensionContext: @unchecked Sendable {}
 
-final class ActionRequestHandler: NSObject, NSExtensionRequestHandling, Sendable {
```
Dimillian, line 18: ?

IceCubesApp/Resources/Localization/fr.lproj/Localizable.strings
```diff
@@ -112,8 +108,6 @@
 "settings.system" = "Paramètres système";
 "settings.content.navigation-title" = "Paramètres de contenu";
 "settings.content.boosts" = "Boosts";
-"settings.content.collapse-long-posts" = "Collapse long posts";
```
Dimillian, line 115: I think there is some problems, you removed a lot of keys instead of translating them?

Issue #1087, opened 2023-02-26
Feature request: Add an indication along left menu of which account is active


mikernc:
I’m sorry if this a duplicate (plus, I cannot a roadmap where this might already exist)…
I have collected 3 different Mastodon accounts that I’ve configured in Ice Cubes. It would be helpful if the profile icon in the left menu of whatever timeline I’m looking at was highlighted in some way (checkmark, highlighted border, <something else>).

flo1o1:
I second that. It would be nice if the avatar of the active account would show up in the sidebar (profile tab) instead of the default user icon.

PR #1159, opened 2023-03-04
Add checkmark for active account

kitwtnb:
I added checkmark for active account.

#1087


Dimillian:
I'm so sad we can't have anything better. I plan to do a bottom sheet with iOS 16.4 to do a proper view. I might wait for it.

Dimillian:
Closing it as we now have a proper account selector

PR #1170, opened 2023-03-05
Traditional Chinese (zh-Hant) localization

sh95014:
Strings localized for Traditional Chinese (Taiwan).


Dimillian:
Awesome! thanks!

PR #1174, opened 2023-03-05
Polish localization

AndrzejRozga:
Fixed problem with with the translation of "mentions" in plural form 
- Localizable.stringsdict needs an integer (not string) to make a choice of the correct plural form.
- fixed StatusEditorAutoCompleteView
- changed %@ format specifier in Localizable.strings with %lld


Dimillian:
Thanks for doing that!

PR #1178, opened 2023-03-05
Update Basque plurals

xabirequejo:
I seem to be unable to get plurals right but I want to believe I'm getting close 😉


Dimillian:
🚀

PR #1187, opened 2023-03-06
Fixed the problem that the header information on the profile screen was not displayed

nekolaboratory:
In my case, the tag search is generally closed, so the user profile screen does not return the list of tags.
However, I noticed that the current implementation has a problem that most of the header information is not displayed on the profile screen for that reason.
Here's how to solve it.

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailViewModel.swift
+++ b/Packages/Account/Sources/Account/AccountDetailViewModel.swift
@@ -129,17 +129,18 @@ class AccountDetailViewModel: ObservableObject, StatusesFetcher {
 
   private func fetchAccountData(accountId: String, client: Client) async throws -> AccountData {
     async let account: Account = client.get(endpoint: Accounts.accounts(id: accountId))
-    let featuredTags: [FeaturedTag]
-    do {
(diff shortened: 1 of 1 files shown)


Packages/Account/Sources/Account/AccountDetailViewModel.swift
```diff
@@ -129,7 +129,12 @@ class AccountDetailViewModel: ObservableObject, StatusesFetcher {
 
   private func fetchAccountData(accountId: String, client: Client) async throws -> AccountData {
     async let account: Account = client.get(endpoint: Accounts.accounts(id: accountId))
-    async let featuredTags: [FeaturedTag] = client.get(endpoint: Accounts.featuredTags(id: accountId))
+    let featuredTags: [FeaturedTag]
+    do {
```
Dimillian, line 133: You can't fix it like that, it'll break the concurrency, well basically it'll first await this call and then do the second one. I'll look to do it another way.

nekolaboratory: Thank you for your code review.
Fixed to a different approach that does not cause errors in processing parallelism.
how about this?

Dimillian: Great thanks!

PR #1199, opened 2023-03-08
Tweak action feedback settings

nekolaboratory:
The sound effects were very good, but it has to be quiet for public use.
So I looked for a menu to turn this feature off, but it was in an unexpected place.
I think it should be displayed without scrolling to get to the desired settings menu more intuitively.

Simulator Screen Shot - iPhone

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/HapticSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/HapticSettingsView.swift
@@ -10,14 +10,24 @@ struct HapticSettingsView: View {
 
   var body: some View {
     Form {
+      if true {
+        Section {
+          Toggle("settings.haptic.timeline", isOn: $userPreferences.hapticTimelineEnabled)
+          Toggle("settings.haptic.tab-selection", isOn: $userPreferences.hapticTabSelectionEnabled)
+          Toggle("settings.haptic.buttons", isOn: $userPreferences.hapticButtonPressEnabled)
+        } header: {
+          Text("Haptic Feedback")
+        }
+        .listRowBackground(theme.primaryBackgroundColor)
+      }
       Section {
-        Toggle("settings.haptic.timeline", isOn: $userPreferences.hapticTimelineEnabled)
-        Toggle("settings.haptic.tab-selection", isOn: $userPreferences.hapticTabSelectionEnabled)
-        Toggle("settings.haptic.buttons", isOn: $userPreferences.hapticButtonPressEnabled)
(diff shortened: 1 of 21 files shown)


nekolaboratory:
This is a PR to check the configuration of the screen, and if it is accepted, I need to clean up the code.
Please feel free to comment on the brush-up of the screen configuration.

te6-in:
What do you think about renaming `Action Feedback Settings` to `Sounds and Haptics` or something? I think the current could feel a bit ambiguous.

Dimillian:
Oh sorry I forgot to comments on that. I'm not a big fan of nesting the setting here and renaming the whole section. I think it's fine at the root? Maybe I'll instead re group the settings in the root screen in different sections.

nekolaboratory:
Thank you for yours comment.
The label name is based on the internal code that will be changed in the future, so there is no particular preference.
At first glance, haptic feedback and sound effects fall into separate categories, but since they are the result of user interaction, we plan to combine them in the future when similar settings are added to the feedback settings page.
Rename the internal code from Haptic Feedback to Action Feedback.
So, I tentatively set the current label name based on the code naming, but if you have an appropriate name, please give me some advice.

nekolaboratory:
> What do you think about renaming `Action Feedback Settings` to `Sounds and Haptics` or something? I think the current could feel a bit ambiguous.

The history of naming is commented at the bottom, but the straight expression may be preferable at present 👍
(Until there are more feedback settings than that)

nekolaboratory:
> Oh sorry I forgot to comments on that. I'm not a big fan of nesting the setting here and renaming the whole section. I think it's fine at the root? Maybe I'll instead re group the settings in the root screen in different sections.

I commented on the history of this PR at the beginning and in the appendix, but I forgot where to turn on and off the sound effects that I translated myself, so I was looking for it in the content setting screen first 😂
It took me a while to find it in the root screen of settings that was hidden by scrolling.
On iPad or Mac, root may be more direct, but on iPhone it's hidden off screen, so it may be a little unfriendly, but I respect the opinion of the main committer.

nekolaboratory:
Simulator Screen Shot - iPhone

Changed the title text.
I hope it's more intuitive and easier to understand. Thanks @te6-in

@Dimillian What do you think of this change?
As an advantage of this screen, I think that you can increase the number of setting items for sound effects and turn them on and off more precisely.
Complete the implementation if necessary.
Feel free to close it if you don't need it :]

Dimillian:
Closing for now. 
Thanks for all the work but I have an idea to rework it a bit later on :)

PR #1228, opened 2023-03-12
Changed the display of numerical values ​​toward accurate values

nekolaboratory:
I see this client as a mastodon tool for advanced users rather than a simple and easy tool for beginners.
Accurate information is therefore preferred over terse numerical representations.
This change was made to comply with the actual values.


Dimillian:
Closing this one. This is a fix I've made recently because it's look nicer for the UI and are more in line with other clients. Imagine when you have 100K followers, it take a lot of space. But I'll find another way to see the full count.

PR #1234, opened 2023-03-13
Changed the display of numerical values ​​toward accurate values

nekolaboratory:
another approach #1228

Changed to display the number of posts in the navigation title.
Simulator Screen Shot - iPhone

### DIFF
--- a/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
@@ -290,6 +290,7 @@
 "account.list.name" = "Назва спісу";
 "account.post.pinned" = "Замацаваны допіс";
 "account.posts" = "Допісы";
+"account.posts.title-%@" = "%@ Допісы";
 "account.relation.follows-you" = "Сочыць за табой";
 "account.joined" = "Далучыліся";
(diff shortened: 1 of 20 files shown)


Dimillian:
This might appear weird as it'll be there even if not scrolled in. I'm with you here that it would be nice to see the full numbers somehow but I'm not yet sure of how.

sothawo:
Does it really make sense for the normal user to have this detailed information? 81,190 posts instead of  80k? Does this make a difference?

Citing from your first PR:

> I see this client as a mastodon tool for advanced users rather than a simple and easy tool for beginners.
Accurate information is therefore preferred over terse numerical representations.

For me this is an app not for advanced users, but for the everyday user and I'm not sure if these detailed numbers make sense for everyone. I'd prefer to have a compact version of these numbers.

nekolaboratory:
> Does it really make sense for the normal user to have this detailed information? 81,190 posts instead of 80k? Does this make a difference?

The official app meets your needs (the display of the number of posts is compact)

> For me this is an app not for advanced users, but for the everyday user and I'm not sure if these detailed numbers make sense for everyone. I'd prefer to have a compact version of these numbers.

What is the difference between IceCubes and the official app?
As I mentioned above, I think that users usually use the official app and it is for advanced users who want a free and high-performance client.

PR #1237, opened 2023-03-13
Add the ability to translate using deepl even if the instance offers its own service

Havhingstor:
I've noticed that the translation service provided by the instance is sometimes not very good and (at least on my instance) does not dynamically detect the language of a post. The user can now force translation in DeepL, e.g. if they notice a problem with the standard translation. Additionally, a user who is very unhappy with the translation service offered by the instance can force translation via DeepL for all posts. For that latter feature, they must provide their own DeepL API key to keep the number of requests on @Dimillian's API key low.


Packages/Env/Sources/Env/KeychainHelper.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+
+public enum KeychainHelper {
```
Dimillian, line 3: We already gave KeychainSwift in the app. I don't think you need this. You can look at how AppAccount is done and maybe use the same library and pattern?

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
```diff
@@ -288,28 +289,54 @@ public class StatusRowViewModel: ObservableObject {
...
+          self.translation = translation
+          isLoadingTranslation = false
+        }
+        
+        return
+      } catch {}
     }
+      
+    // If not or fail we use Ice Cubes own DeepL client.
+    await translateWithDeepL(userLang: userLang, userAPIKey: userAPIKey)
```
Dimillian, line 311: Could this also be called in the catch like before? If the user don't set their own API Key it'll use mine right? So this should be fine like before.

Packages/Env/Sources/Env/DeepLUserAPIHandler.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import SwiftUI
+
+@MainActor
+public enum DeepLUserAPIHandler {
```
Dimillian, line 5: I find this a little bit hard to read, I wonder if it can be cleaned when using KeychainSwift.

Dimillian:
This is really awesome! I've made a few fixed to the theme etc.. and added a few comments. It could probably be simplified a bit.

Packages/Status/Sources/Status/Row/StatusRowViewModel.swift
Havhingstor: It'll use yours. 
I don't think it's needed to call translateWithDeepL in the catch block since we already call that function after we finished the if statement (if the return statement doesn't get executed). If the request to the instance fails, the empty catch is executed, then the code after the catch, which includes a call to translateWithDeepL (that call is only not executed if the return statement in the if statement (ln 306) is reached, which happens if the instance request is successful).

Issue #1239, opened 2023-03-13
Feature request: Reordering of Remote Local Timelines


gsax:
Allow the reordering of the added Remote Local Timelines in the options.

PR #1240, opened 2023-03-13
Allow reordering of remote timelines

Havhingstor:
The reordering of remote timelines is now available. See #1239


Dimillian:
This is nice thanks!

PR #1248, opened 2023-03-14
Fix the translate-with-deepl context button

Havhingstor:
Previously, the button was shown if the always use DeepL setting was active. This is redundant.

There are two possible fixes:
- We could keep that button if the user hasn't selected always-use-DeepL (we would therefore use the standard API key). The user could use it if he's generally satisfied with the instance-translation service, but needs to use DeepL for one specific post.
- We could do the same thing, but show the button only if the user has saved his own API key, (I would then make it possible to set a key without activating always-use-DeepL). It would work the same and have the same effect but would keep traffic on @Dimillian's key lower
- We could (theoretically, I don't really think that's necessary) show the button if the always-use-DeepL option is selected, which would change the behavior of the normal translate button in the context menu to use the instance service.

I would vote for one of the first two options, @Dimillian should decide if he wants the button to run via his API key.

Theoretically, I could also implement (1 or 2) AND 3, which would allow the user to always (at lest with a saved key) switch the translation service.

### DIFF
--- a/Packages/Status/Sources/Status/Row/Subviews/StatusRowContextMenu.swift
+++ b/Packages/Status/Sources/Status/Row/Subviews/StatusRowContextMenu.swift
@@ -138,7 +138,7 @@ struct StatusRowContextMenu: View {
         Label("status.action.translate", systemImage: "captions.bubble")
       }
       
-      if viewModel.alwaysTranslateWithDeepl {
+      if !viewModel.alwaysTranslateWithDeepl {
         Button {
           Task {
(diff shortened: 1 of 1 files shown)


Dimillian:
I think the first option is the right one. When I inverted this case I thought it would not work but yes indeed it's working even without the settings active as it use my key. 
But it indeed look weird to always show two translates button.

Havhingstor:
Then this PR should enable the first version. I'd keep it that way since e.g. my instance doesn't automatically detect the post language, which leads to problems because unfortunately there still are wrongly classified posts.

PR #1258, opened 2023-03-17
Settings screen Accessibility tweaks

nesevis:
# This PR

1. Removes `button` traits from the settings options that are `Link`-backed. 
 
These two traits indicate different behaviours and shouldn't co-exist on the same element. Currently, SwiftUI sets both, which is being tracked as _FB11507660_.

2. Wraps `AppAccountView` in a `Button`, which results in a single accessibility element per account. 

Previously, the selected tick would render as its own element, as would the profile image (with no descriptive label). Because there was a `onTapGesture` attached to a `.contentShape`, each of these 5 component elements would be interactive, all performing the same action.

## Before
IceCubes-Settings-Before

## After
IceCubes-Settings-After


Dimillian:
For the account button, it make the label using the tint color instead of being labelColor. I think it should be explicitly labelColor then.

Dimillian:
Fixed. Thanks for this!

PR #1259, opened 2023-03-17
Compose Post Screen Accessibility Tweaks

nesevis:
# This PR

… makes a few relatively small changes that make big improvements to the accessibility of the screen:

- Adds a localized label for the AI prompt accessory view icon. Previously this was falling back to the SF symbol key _faxmachine_.
- Darken the remaining character count element to a `.secondary` colour, which just passes [APCA Bronze](https://www.myndex.com/APCA/), but is still quite a way off passing WCAG 2.1 AA (3.3:1 vs 4.5:1). 
- Changes the remaining character count element to have the `.updatesFrequently` trait, moves the count to the `accessibilityValue` and adds a label that contextualises the value.
- Sets `accessibilitySortPriority` on the contents of the `ScrollView`. Previously, the elements in the scroll view would come last in the traversal order, meaning that a VoiceOver user would move from the disabled _Post_ button in the nav bar to the first of the accessory view icons.
- Changes the `TextView` accessible behaviour to match that of other text inputs: On most text fields, the place holder value is set as the `accessibilityValue` before the user enters text, after which point it is replaced. I also set `accessibilityHidden` on the separate placeholder view, which you can see was also part of the accessibility tree.
- Hides the `AvatarView` from accessibility. There is no functionality associated with it on this screen, so it is effectively decorative. There is no need to display it, even with alt text, as this is the user's own avatar (which they would presumably be familiar with 😅 )
- Changes the `privacyMenu`, moving the current value to `accessibilityValue`, and adding context through a new label and hint.
- Adds `.button` trait and the emoji shortcote as `accessibilityLabel` in the custom emoji picker, which previously offered no way for VoiceOver users to discern what the emoji was.
- Changes the list items in the drafts sheet to be buttons, rather than `Text` + tap gesture.

<small>_All screenshots are from [Reveal](https://www.revealapp.com/)_</small>

## Compose screen, before
The compose screen VoiceOver element breakdown before the changes in this PR

## Compose screen, After
The compose screen VoiceOver element breakdown after the changes in this PR

## Custom emoji sheet, before

The compose custom emoji sheet VoiceOver element breakdown before the changes in this PR

## Custom emoji sheet, after

The compose custom emoji sheet VoiceOver element breakdown after the changes in this PR

### DIFF
--- a/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
@@ -519,6 +519,8 @@
 "accessibility.editor.button.language" = "Мова";
 "accessibility.editor.button.ai-prompt" = "AI prompt";
 "accessibility.editor.button.characters-remaining" = "Characters remaining";
+"accessibility.editor.privacy.label" = "Visibility";
+"accessibility.editor.privacy.hint" = "Changes post audience.";
 "accessibility.tabs.timeline.add-account" = "Дадаць уліковы запіс";
 "accessibility.app-account.selector.accounts" = "Уліковыя запісы";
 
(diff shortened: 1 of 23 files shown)


Dimillian:
This is amazing!!

Dimillian:
All good! Merging.

PR #1274, opened 2023-03-18
Profile tab accessibility uplift

nesevis:
# This PR

…is a major one. There were a fair few changes here that affect the accessible representation in other parts of the app, most notably the non-focused — not displayed as a single post — `StatusRowView`.

_Touch target sizes remains a concern — a lot of the elements are less than the recommended 44x44pts, but changing them would necessitate substantial design changes, so I've left them off._

## Profile tab specific

- Combines many partial views into one accessibility element
- Adds labels and hints to the header and avatar images
- Tweaks the order of elements to flow more naturally
- Tweaks `CustomInfoLabels` to have the numerical value as its `accessibilityValue`, and adds hints indicating what action they have (scrolling or navigating)
- Changes the `accessibilityLabel` of the nav bar item to _Options_ and adds more alternatives to allow for faster use with Voice Control and Full Keyboard Access
- Adds explicit `accessibilityLabel`s to the `Picker`. Previously, these were the fallback labels provided by SF Symbols, which didn't really make complete sense.
- Removes `StatusRowView` swipe actions if VoiceOver is running. These actions were folded into the custom actions on the accessibility element, which meant there were a lot of duplicate actions being surfaced.
- Removes user interaction flag from `staticText` elements with no associated action
- Changes the accessible representation of `FollowButton`s to a `Toggle`, which is the natural interaction mode for these. This means that VoiceOver will read _Display boosts, Switch Button, on, Double-tap to toggle setting._, which provides ample context.
- Combines the user-defined fields title and values, and adds a _Verified_ `value` where applicable. For valid URLs, a `.link` trait is added automatically. The group of user-defined fields is now its own accessibility container, so when a user moves focus to the container, it will prefix _User-defined fields_ before reading the element description.
- Add image descriptions for familiar follows to their `safeDisplayName`; previously there was no way to know the difference between them. Also added a `.button` trait, as tapping them performs navigation.

## More general
- Adds explicit `accessibilityLabel`s to `StatusRowActionsView` actions. Previously, they were _close_, but there was one SF symbol key that snuck in there.
- Adds image alt text to the `StatusRowView` `accessibilityLabel`. If there is more than one image present, it will read the first one and indicate that there is more alt text available.
- Create an explicit accessibilityLabel for unfocused `StatusRowView`s.
- Fix a bug that prevented people with VoiceOver enabled from double-tapping a `StatusRowView` to go to the post detail.

## `StatusRowView` combined label

This is a bit of a complicated one ([code](https://github.com/Dimillian/IceCubesApp/blob/61afd769fd55e2f4e825f9e0ba6f9206f7b001b4/Packages/Status/Sources/Status/Row/StatusRowView.swift#L205-L298), reuses existing logic where possible), but the end result looks like this for a plain post with one image with alt text:

> [RobWhitaker new] I just connected a Bluetooth keyboard to my iPhone. Can anyone explain how connecting a keyboard can affect Wi-Fi?  **Image alt text: Using "keyboard" may affect Wi-Fi and Bluetooth connectivity.** 1 wk. ago, 3 replies, 1 boosts, 0 favorites

For a boosted post:

> Patrick H. Lauke boosted Frank Whitehouse Today I finally told my kids that St Patrick isn't real, and it's been me putting the snakes under their pillows all these years.  1 day ago, 8 replies, 218 boosts, 12 favorites

For a reply:

> Ellen Shapiro replied to @malin Oh those came out nice! What kind of wood/stains did you use?   9 hr. ago, 0 replies, 0 boosts, 0 favorites

# Profile tab, before

Profile-Tab-VoiceOver-Before

Profile-Tab-VoiceControl-Before

# Profile tab, after

Profile-Tab-VoiceOver-After

Note that for the user-defined fields, VoiceOver on device will add `link` traits where it detects urls; adding them in SwiftUI will result in `Link` being read out twice.

Note also that  the _Following_, _Receive notifications_ and _Display boosts_ will be read as _Label, Switch Button, on|off_, but Reveal doesn't support the private `toggle` trait Apple uses for that label yet.

Profile-Tab-VoiceControl-After

# Timeline, before

Timeline-VoiceOver-Before

_Note the "0,0,0, Button" label, which I believe is some kind of bug that inserted the full description into a custom action label_

# Timeline, after

Timeline-VoiceOver-After

### DIFF
--- a/Packages/Status/Sources/Status/Row/StatusRowView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowView.swift
@@ -248,12 +248,12 @@ private struct ConditionalAccessibilityLabelModifier: ViewModifier {
     Text(hasSpoiler
       ? viewModel.finalStatus.spoilerText.asRawText
       : viewModel.finalStatus.content.asRawText
-    ) + Text(" ") +
+    ) + Text(", ") +
     Text(hasSpoiler
       ? "status.editor.spoiler"
       : ""
-    ) + Text(" ") +
-    imageAltText() + Text(" ") +
+    ) + Text(", ") +
+    imageAltText() + Text(", ") +
     Text(viewModel.finalStatus.createdAt.relativeFormatted) + Text(", ") +
     Text("status.summary.n-replies \(viewModel.finalStatus.repliesCount)") + Text(", ") +
     Text("status.summary.n-boosts \(viewModel.finalStatus.reblogsCount)") + Text(", ") +
@@ -265,9 +265,9 @@ private struct ConditionalAccessibilityLabelModifier: ViewModifier {
       case (true, false):
         return Text("accessibility.status.a-replied-to-\(finalUserDisplayName())") + Text(" ")
       case (_, true):
-        return Text("accessibility.status.a-boosted-b-\(userDisplayName())-\(finalUserDisplayName())")  + Text(" ")
+        return Text("accessibility.status.a-boosted-b-\(userDisplayName())-\(finalUserDisplayName())")  + Text(", ")
       default:
-        return Text(userDisplayName()) + Text(" ")
+        return Text(userDisplayName()) + Text(", ")
     }
   }
 
--- a/Packages/Status/Sources/Status/Row/StatusRowView.swift
+++ b/Packages/Status/Sources/Status/Row/StatusRowView.swift
@@ -216,7 +216,6 @@ private struct ConditionalAccessibilityLabelModifier: ViewModifier {
     viewModel.displaySpoiler && viewModel.finalStatus.spoilerText.asRawText.isEmpty == false
   }
 
-  // viewModel.finalStatus.spoilerText.asRawText
   func body(content: Content) -> some View {
     if setLabel {
       if hasSpoiler {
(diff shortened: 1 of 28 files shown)


Dimillian:
I'll probably don't have time to look at it this weekend but for anything touching the status row would be good to have some performances regressions tests. Just to be sure it's not getting slower as the timeline scroll.

Dimillian:
But this is amazing!!

Packages/Status/Sources/Status/Row/StatusRowView.swift
```diff
@@ -203,3 +201,98 @@ public struct StatusRowView: View {
     .transition(.opacity)
   }
 }
+
+/// A ``ViewModifier`` that creates a suitable combined accessibility label for a `StatusRowView` that is not focused.
+private struct ConditionalAccessibilityLabelModifier: ViewModifier {
```
Dimillian, line 206: Profiled this, this is fine. I can definitely see it in the body timing profiler but nothing heavy. Too bad it's computed even if accessibility features turned off.

Dimillian (approved):
All good. Tested + review the code, this is awesome!

Issue #1267, opened 2023-03-18
Feature request:translate profile info


paul4565:
Search that this request don't already exist before creating it.

paul4565:
Many thanks 👍

Dimillian:
Thanks @Havhingstor :)

PR #1276, opened 2023-03-18
Allow translation of an account bio/note

Havhingstor:
The bio (note) of an account can now be translated via DeepL. If the user has put in his own DeepL API key, that is used, otherwise, the standard one is used. See #1267


Dimillian (approved):
Super nice thanks!

PR #1278, opened 2023-03-19
Fix: Inline StatusRowView accessibilityLabel modifier to avoid flashing

nesevis:
By inlining this label, we avoid the creation of `ConditionalContent`, which often leads to views being recreated unnecessarily due to identity issues.

In focused mode, the empty label is not read as it is the accessibility container label for the post component elements.


Dimillian:
It's indeed fix it, thanks!

PR #1279, opened 2023-03-19
de: translated the whole lot of new strings

dertuxmalwieder:


Dimillian:
And there is more coming thanks @nesevis 😆 
Thanks!

PR #1288, opened 2023-03-21
Update German localization

J-rg:


Havhingstor:
Der [Duden](https://www.duden.de/konjugation/stellen) sagt, dass wohl beides geht.

dertuxmalwieder:
Personally, I'd use "stell" (etc.), but I won't put my veto (thanks...) on "stelle" (etc).

PR #1292, opened 2023-03-23
Accessibility tweaks + Notifications and Messages tab uplift

nesevis:
# This PR

… is relatively small, all things considered. The fixes here are mostly iterative, but represent big wins.

## Timeline / Post composition / `StatusRowView`

- Adds links and hashtags in post content to the Actions rotor. This means that users can interact with all the content in the post.
- Adds reply and quote options to the Actions rotor. These were left off by mistake in the previous round of accessibility work.
- Adds haptics to all `StatusRowView` actions
- Removes all `accessibilityActions` from the `StatusRowView` if `viewModel.showsActions` is false.
- Fixes a bug in the `accessibleRepresentation` of the toolbar title. Previously, it would have a `.button` trait and indicate that it would open a pop-up menu regardless of whether this was actually true. This primarily affected the child screens of the timeline tab.
- Removes the `accessibilityCustomContent` representation of the full post content. This was causing it to be read out twice in the detail view.
- Combines tag header view elements
- Combines quote/reply composition screen `EmbeddedStatusView` elements and hides actions.

### Before
Hashtag-Timeline-Before

Toot-Composition-Before

### After
Hashtag-Timeline-After

Toot-Composition-After

## Notifications

- Hides `topPaddingView` from accessibility. Its job is purely visual and doesn't need to be presented to VoiceOver.
- Combines `NotificationRowView` child elements into one element
- Adds all account references in `.follow`-type notifications to the Actions rotor so they can be interacted with.

### Before
Notifications-Before

### After
Notifications-After

## Conversations

- Combines `ConversationListRow` child elements into one element.
- Adds magic tap VoiceOver interaction that opens the reply status editor. The default action is to push into the conversation view.
- Fleshes out the Action rotor with actions that correspond to the context menu, plus extras for user, link and hashtag references.
- Adds an accessible representation of the unread status indicator. The first string in the `accessibilityLabel` will now be _Unread_ if this is the case.

### Before
Messages-Before

### After
Messages-After


Dimillian:
Awesome, thanks!

PR #1307, opened 2023-03-25

Jerry23011:
### Changes
- Added new translations

@nixzhu could you take a look :)


nixzhu:
LGTM

PR #1310, opened 2023-03-26
Fix icon switching on iPad

Havhingstor:
The switching of icons didn't work on the iPad anymore (See https://norden.social/@betamax65/110088664994961476). This might be a temporary fix, depending on what caused the bug (which I couldn't determine). For now, one icon in the set seemingly must have a complete iPad icon set, which is achieved with the standard icon. Furthermore, an assert statement helps to determine the problem with the icons. This statement won't be compiled into release builds.


Dimillian:
Working fine for me. I think it's just and iOS bug. Just need to reboot. We don't need full icon set. I don't think it's broken.

Dimillian:
So, I get it to reproduce once in a while, on an iPad 16.4 hmmm

Dimillian:
It's indeed fixing it. This is crazy. But ok.

PR #1342, opened 2023-03-31
Fix the lag of the display setting sliders

### LINKED ISSUE #1341: Bug: laggy sliders
- OS: 
 iOS 16.4
 - IceCubesApp version: 1.6.12

## Description

The sliders for customizing font size and line spacing are very slow. I don’t know if it’s a swift-ui problem but you should consider to use some different UI component if you can’t fix them

image

Havhingstor:
The sliders in the display settings were laggy because changing the value in the theme class needs comparatively much time. This writing to the class is now only done when the user lets go of the slider, so sliding it is responsive. I was unable to make writing to the class more responsive, the example post needs therefore a short time before it accepts the new values. Furthermore, all AppStorage keys are now realized with a ThemeKey to clean the class up. Fixes #1341

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/DisplaySettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/DisplaySettingsView.swift
@@ -6,11 +6,13 @@ import Network
 import Status
 import SwiftUI
 
-class DisplaySettingsLocalColors: ObservableObject {
+class DisplaySettingsLocalValues: ObservableObject {
   @Published var tintColor = Theme.shared.tintColor
   @Published var primaryBackgroundColor = Theme.shared.primaryBackgroundColor
   @Published var secondaryBackgroundColor = Theme.shared.secondaryBackgroundColor
   @Published var labelColor = Theme.shared.labelColor
+  @Published var lineSpacing = Theme.shared.lineSpacing
+  @Published var fontSizeScale = Theme.shared.fontSizeScale
 
   private var subscriptions = Set<AnyCancellable>()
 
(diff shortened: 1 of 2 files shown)


IceCubesApp/App/Tabs/Settings/DisplaySettingsView.swift
```diff
@@ -147,22 +149,36 @@ struct DisplaySettingsView: View {
       .navigationDestination(isPresented: $isFontSelectorPresented, destination: { FontPicker() })
 
       VStack {
-        Slider(value: $theme.fontSizeScale, in: 0.5 ... 1.5, step: 0.1)
-        Text("settings.display.font.scaling-\(String(format: "%.1f", theme.fontSizeScale))")
+        Slider(value: $fontScale, in: 0.5 ... 1.5, step: 0.1) { editing in
```
Dimillian, line 152: I think this should be done like I've solved lagging for colors. 
Look at `DisplaySettingsLocalColors`
Probably rename it to DisplaySettingsLocalValues.
And connect the slider values the same way. This way we debounce them and the slider is not laggy.

Packages/DesignSystem/Sources/DesignSystem/Theme.swift
```diff
@@ -129,8 +130,8 @@ public class Theme: ObservableObject {
   @AppStorage(ThemeKey.followSystemColorSchme.rawValue) public var followSystemColorScheme: Bool = true
   @AppStorage(ThemeKey.displayFullUsernameTimeline.rawValue) public var displayFullUsername: Bool = true
   @AppStorage(ThemeKey.lineSpacing.rawValue) public var lineSpacing: Double = 0.8
-  @AppStorage("font_size_scale") public var fontSizeScale: Double = 1
-  @AppStorage("chosen_font") public private(set) var chosenFontData: Data?
+  @AppStorage(ThemeKey.fontSizeScale.rawValue) public var fontSizeScale: Double = 1
```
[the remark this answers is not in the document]
Havhingstor: Removed it since it's not necessary

IceCubesApp/App/Tabs/Settings/DisplaySettingsView.swift
Havhingstor: Implemented it

Dimillian: Perfect thanks!

PR #1347, opened 2023-04-01
Restore HTMLString link parsing

nesevis:
This fix for d3888d8c  piggybacks on the `SwiftSoup` parser and does not operate on the `AttributedString` at all.

edit: It should have no performance overhead.


Dimillian:
Thanks!

PR #1364, opened 2023-04-07
Show verified URLs in account lists

### LINKED ISSUE #1361: Feature request: Indicate verified URLs in search results (as mentioned on ATP)
In the ATP podcast, they mentioned that it would be useful if Mastodon client search results showed the verified URL(s) of accounts so you can tell quickly which one is the official account.

It’s a neat idea. WDYT? Is it technically feasible without N+1 network requests (maybe that already happens and is OK)?

Havhingstor:
This allows the user to quickly assess which account of multiple is the official one, especially when searching for a person with multiple search results. Fixes #1361


Dimillian:
Love it a lot, thanks!

PR #1410, opened 2023-05-09
Add option to not show replies

### LINKED ISSUE #1209: Feature request: Allow hiding replies from timeline
Browsing your timeline shows replies which are often nonsensical shown out of context. Not to mention that being in reverse chronological order makes them even more unintelligible.

The option to hide them (showing only the first post in the thread with a reply count indicator) would let users decide if they want to click/tap on it to display all the posts in the conversation from oldest to newest.

Havhingstor:
An user option to not show replies. There is the possibility to overwrite the option via a flag set when creating the view. This is necessary to show the replies when being in the specific tab in the account view. This option is triggered for all statuses in the account views, the display of replies is there (account tab) only managed via the old way.
Closes #1209

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
@@ -19,6 +19,13 @@ struct ContentSettingsView: View {
         }
       }.listRowBackground(theme.primaryBackgroundColor)
 
+      Section("settings.content.replies") {
+        Toggle(isOn: $userPreferences.showReplies) {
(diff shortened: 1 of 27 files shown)


vishalMalvi (approved):

wnagrodzki:
@Havhingstor Hi, is any help needed for getting that through?

Havhingstor:
@Dimillian said he wanted to take a close look since changes to PendingStatusesObserverView might introduce performance problems: https://mstdn.social/@havhingstor/110632986579085701. I’m gonna fix the current merge conflicts soon, I don’t think they’re a problem.

PR #1421, opened 2023-05-29
Add setting to control share button default behavior

nathreed:
This adds a setting to control the behavior of the share button on the status row actions view. Currently, it always shares the link to the post as well as the post text. 

In iOS 16.4, Apple added iMessage unfurling for Mastodon URLs. When sharing posts from Ice Cubes via iMessage, this leads to the recipient seeing two copies of the post: one from the unfurled link and one from Ice Cubes including the post text. 

Users will now have the option to exclude the post text from their sharing. This is easier than tapping the 3-dots button on the post (which is kind of small) and then expanding the Share menu in the context menu, which is the other way to access this functionality at the moment.

The default value for the new option will be "Link and Text", which is the current behavior - so we won't change the behavior on existing users.

[Settings Screenshot](https://github.com/Dimillian/IceCubesApp/assets/1245988/1d3070d0-da2c-4728-8007-d7742f0c1fe8)

### DIFF
--- a/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
@@ -130,6 +130,10 @@
 "settings.content.media.show.alt" = "Show ALT Texts";
 "settings.content.reading" = "Reading";
 "settings.content.posting" = "Posting";
+"settings.content.sharing" = "Sharing";
+"settings.content.sharing.share-button-behavior" = "Share Button Behavior";
+"settings.content.sharing.share-behavior.link-only" = "Link Only";
+"settings.content.sharing.share-behavior.link-and-text" = "Link and Text";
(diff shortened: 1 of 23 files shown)


IceCubesApp/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -130,6 +130,10 @@
 "settings.content.media.show.alt" = "Show ALT Texts";
 "settings.content.reading" = "Reading";
 "settings.content.posting" = "Posting";
+"settings.content.sharing" = "Sharing";
```
Dimillian, line 133: Those strings need to be added to every localization files. You can leave them in English as it'll be translated later. Thanks!

nathreed: addressed in b95741b

PR #1442, opened 2023-06-26
More compact account list rows

Dean151:
Current issue:
- The account list row is not compact; it shows the bio entirely, that might take an entire screen!
- followers count is often a relevant information to get on that screen, to decide to follow back, or simply to see the impact of a boost.

Proposal:
- Followers count & verification link before bio
- bio limited to 2 lines
- Using footnote for the verified site ; and caption for the bio 

Before / After images
Capture d’écran 2023-06-26 à 13 57 50 Capture d’écran 2023-06-26 à 13 57 17


Dimillian:
Very nice!

PR #1453, opened 2023-06-30

Jerry23011:
### Changes
- Add new translations for Sharing settings
- Fix translations in Display settings


nixzhu:
LGTM

Issue #616, opened 2023-02-02
Feature request: Mute account on post


kylebavis:
It would be useful to have the option to mute an account directly from hamburger menu on a post. Currently you need to tap through to the profile and select it there.

kylebavis:
I got a lot of use out of the analogous feature on twitter, mostly in cases where I followed someone who posts things I want to see but sometimes boosted people I wasn’t interested in. Right now pretty much all mastodon clients require you to temp through to the profile to mute, so my net number of taps for timeline curation is up substantially. I curate my feed pretty aggressively between account mutes and keyword filters. 

Admittedly, this might be an unusual pattern.

alanmartello:
I agree that mute here, especially timed mute, would be handy on the timeline. I often have a few hundred unread items and quick timed muting (ex. mute 3 hours) here would allow triaging / prioritizing this list from the list itself. For me this is not about muting an account forever it's about quickly removing things from the timeline because I want to ignore this account right now.

Doing it here would save me both the mental context switch and the time to tag the remote instance to load the profile of the user. No need for me to see the profile, I already know what I want to do.

RobATX:
Maybe consider making press-hold on profile icon options different from menu ... options.

mattrrs:
I agree that it should be on the main feed. If I'm looking at something that's offensive to me but is just fine by the community guidelines, I just want to block it immediately without having to drill down into the profile and see even more stuff that's offensive to me. Accidentally blocking someone is an edge case. And there should be a place to see blocked accounts anyway, in case you want to unblock someone from (e.g.) months ago.

Triffen:
Yes, came to post the same future request, mute and block are expected as they are present both in web interface and Metatext. I don't really know why that menu is so long. Muting, blocking and filtering are the only things I do from that menu on web interface.
As said above, I don't want to visit profiles of the authors of offensive content, I just want it gone immediately.

PR #1460, opened 2023-07-02
Add feature to block or mute user directly from post

nathreed:
Inspired by a user commenting that the app doesn't have this feature: https://ioc.exchange/@shac/110645300514488847

To avoid calling the /accounts/relationships endpoint for every single status displayed, the relationships data is only loaded when the menu is activated.
When the API call comes back, the items are added to the menu (updating the view model appears to cause the menu to update, even while it is displayed)
Borrowed blocking & muting logic/menu items from AccountDetailContextMenu.

Addresses #616


Dimillian:
That's really nice!

PR #1491, opened 2023-07-09
Improved remote timeline adding interface

zhouqimark:
In order to make it possible to add remote timelines multiple times in the same logical interface and to make the user interface more user-friendly


Dimillian:
Closing this one. The UX is not fitting to me, it feels clunky. It also would be the first time we use a toast in the app and I don't really want to introduce this pattern. Thanks!

Issue #1370, opened 2023-04-09
Feature request: Antennas, aka, Let users create "lists/folders" for hashtags and local timelines


aquariusdev:
Building on top of my [previous request](https://github.com/Dimillian/IceCubesApp/issues/541), I propose an idea for aggregating multiple hashtags and remote local timelines into "lists/folders". 

Hashtag follows, at the very least, can fill up quickly. Also, there are several niche instances for a broad range of community types. Being able to add more than one hashtag or instance to a "list" (i.e., Politics) would allow me to organize them better, in the same way that we already do with follow lists. 

This would be akin to how Misskey/Calckey use their antennas feature. This would also build a bit on Mastodon desktop's hashtag pinning feature, which only allows for up to five hashtags to add/filter by default.

Let me know if you need a mockup, I can make a few in GIMP.

Issue #1125, opened 2023-03-01
Feature request:Add ability to search for a hashtag and save to the Explore tab, but as a saved search not a follow. This means it doesn't appear in the Home timeline. Then please enable that hashtag to appear in the timelines drop-down selection. This means a user can easily "dip into" a hashtag occasionally.


ghost:
Search that this request don't already exist before creating it.

paul4565:
This feature is available in Tusker by the way.

Dimillian:
But every apps have different features right :)

paul4565:
Yes true.. it's a good one!

paul4565:
#1214 (recent searches) would be simpler than this and probably easier to make use of.

Issue #541, opened 2023-01-30
Feature request: Allow for followed hashtags to include/exclude additional hashtags


aquariusdev:
Mastodon desktop allows for this. Perhaps like this:
* after the tag has been followed, an edit button appears next to the hashtag on the top bar next to the "Following" button
* the edit button shows the following
* * include additional tags for this search (switch)
* * any of these [field]
* * all of these [field]
* * none of these [field]
* * local only (switch)

aquariusdev:
Explanatory image: ![image](https://user-images.githubusercontent.com/28520002/218416629-e00378d0-1578-4a2a-9287-509054686c63.png)

PR #1506, opened 2023-07-18
Tag groups

alexito4:
Hello, first PR on the project so bare with me ^^

## Overview

This is my take on a timeline of hashtags. I find myself needing this for the way I consume Mastodon (see [post](https://mastodon.social/@alexito4/110711840210384969)) so I decided to take a stab at it after I discovered that the API is capable of it and the advanced web UI has support for it (with a quite obscure UX) (see [screenshot](https://mastodon.social/@alexito4/110734184528755483)) . 

For reference the API endpoing is the same as the one already used for tags. It has query parameters to include "additional" tags in the search. https://docs.joinmastodon.org/methods/timelines/#query-parameters-1 

So this PR adds:
- The concept of Tag Group as a Swift struct.
- Support for the query params in the Endpoint
- A new entry in the Timeline Menu for tag groups. 
- The ability to add a new tag group. 
- The ability to remove and sort tag groups. 

The Tag Group needs at least 2 tags, otherwise is not a group ¯\_(ツ)_/¯  I've heard people say there is a limit of 4 but the API doesn't seem to complain if you send more so I haven't added an upper limit. 

That said something else I wanted to look at was a way to have tag timelines easily accessible WITHOUT following the tag. I don't want to follow tags, they clutter my timeline, but I want to have a quick access to my fav ones. Tag Groups solve this scenario for me but I could see people wanting the same quick access for an individual tag so maybe is worth thinking if this feature could be tweaked a bit to cover both cases. 

Note also that the API has support for other forms of additional tags (all, none), I've just added support for ANY for simplicity and because that seems the common scenario (and the want I need 😜)

Related:
- #1370
- #541
- #1125

## Screenshots

<table>
<tr>
 <td>Menu</td>
 <td>Timeline</td>
 <td>Edit groups</td>
 <td>Settings</td>
</tr>
<tr>
 <td></td>
 <td></td>
 <td></td>
 <td></td>
</tr>
</table>

## Open questions

- [ ] Formatting in the project seems non(Xcode)-standard. Do you have any format automation in place? What's the advice to adhere to the formatting rules?
- [ ] I've added support for SFSymbols in the Tag Group. Ideally it would be nice to have a picker but I just quickly implemented it based on the name of the symbol. I really want my Swift list with the swift icon ^^ But I'm open to removing this if it's not good enough, or find some third party package that offers a nice picker. 
- [ ] Localization: what's the requirement here? I've done the languages I know but not sure what to do with the rest. 
- [ ] a couple of TODOs with ? in the code

### DIFF
--- a/IceCubesApp/App/AppRouter.swift
+++ b/IceCubesApp/App/AppRouter.swift
@@ -32,6 +32,7 @@ extension View {
         ConversationDetailView(conversation: conversation)
       case let .hashTag(tag, accountId):
         TimelineView(timeline: .constant(.hashtag(tag: tag, accountId: accountId)), scrollToTopSignal: .constant(0), canFilterTimeline: false)
+      // TODO: anything to do in the app router????
       case let .list(list):
         TimelineView(timeline: .constant(.list(list: list)), scrollToTopSignal: .constant(0), canFilterTimeline: false)
       case let .following(id):
@@ -82,6 +83,9 @@ extension View {
       case .addRemoteLocalTimeline:
         AddRemoteTimelineView()
           .withEnvironments()
+      case .addTagGroup:
+          AddTagGroupView()
+              .withEnvironments()
       case let .statusEditHistory(status):
         StatusEditHistoryView(statusId: status)
           .withEnvironments()
--- a/IceCubesApp/App/AppRouter.swift
+++ b/IceCubesApp/App/AppRouter.swift
@@ -32,7 +32,6 @@ extension View {
         ConversationDetailView(conversation: conversation)
       case let .hashTag(tag, accountId):
         TimelineView(timeline: .constant(.hashtag(tag: tag, accountId: accountId)), scrollToTopSignal: .constant(0), canFilterTimeline: false)
-      // TODO: anything to do in the app router????
       case let .list(list):
         TimelineView(timeline: .constant(.list(list: list)), scrollToTopSignal: .constant(0), canFilterTimeline: false)
       case let .following(id):
--- a/Packages/Timeline/Sources/Timeline/TimelineView.swift
+++ b/Packages/Timeline/Sources/Timeline/TimelineView.swift
@@ -39,7 +39,9 @@ public struct TimelineView: View {
     ScrollViewReader { proxy in
       ZStack(alignment: .top) {
         List {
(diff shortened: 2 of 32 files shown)


Dimillian:
This is great!
- I'm running SwiftLint myself once in a while so don't worry about it. The only thing that is different from a standard / default Xcode settings is 2 tab instead of 4. 
- Will try it out and come back on you. 
- You need to have the default key and default english value to all files. Contributors who keep track of the localizations will translate it later.
- Will look at the code :)

IceCubesApp/App/AppRouter.swift
```diff
@@ -32,6 +32,7 @@ extension View {
         ConversationDetailView(conversation: conversation)
       case let .hashTag(tag, accountId):
         TimelineView(timeline: .constant(.hashtag(tag: tag, accountId: accountId)), scrollToTopSignal: .constant(0), canFilterTimeline: false)
+      // TODO: anything to do in the app router????
```
Dimillian, line 35: You added your route so it seems all good. This todo can be deleted?

Packages/Timeline/Sources/Timeline/TimelineView.swift
```diff
@@ -180,40 +182,64 @@ public struct TimelineView: View {
...
+                    .font(.scaledHeadline)
+                }
+                .accessibilityElement(children: .combine)
+              }
+          }
+      }
+    }
+    
+    @ViewBuilder
+    private func headerView(
```
Dimillian, line 228: Really nice!

Dimillian:
It's working good, I'll make a push on the PR / cleanup a bit and then merge and take it from there. This is really really good & solid! I just love it and every users will love it! Thank you a lot!

Dimillian:
I think the next todo if you're up to it is to make the add screen works with editing too :)

IceCubesApp/App/AppRouter.swift
alexito4: @Dimillian yes I added a route for the "add screen" but I was not sure if a route was needed for the "tag group timeline" itself.

alexito4:
> I think the next todo if you're up to it is to make the add screen works with editing too :)

Yes that's a missing part, I will see if I have some time soon. 

Cheers

PR #1509, opened 2023-07-19
Allow specifying the default reply visibility

### LINKED ISSUE #1490: Feature request: Add a setting for default replies visibility
I would appreciate an option to set the default visibility for replies to „unlisted“ while new posts remain e.g. „Public“

Havhingstor:
Replies can now have their own default visibility. This visibility is always at
least as restrictive as the default post visibility. When posting a reply, the
visibility is pre-populated with the more restrictive out of the default visibility and
the visibility of the original post.
Closes #1490

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
@@ -110,7 +110,7 @@ struct ContentSettingsView: View {
   }
 }
 
-extension View {
+fileprivate extension View {
   func onChange(of: some Equatable, _ perform: @escaping () -> Void) -> some View {
     if #available(iOS 17, macOS 14, *) {
       return onChange(of: of) { oldValue, newValue in
--- a/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
+++ b/IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
@@ -92,7 +92,7 @@ struct ContentSettingsView: View {
             }
           }
         }
-        .onChange(of: userPreferences.postVisibility) { newValue in
+        .onChange(of: userPreferences.postVisibility) {
           userPreferences.conformReplyVisibilityConstraints()
         }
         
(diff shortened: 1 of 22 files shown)


IceCubesApp/App/Tabs/Settings/ContentSettingsView.swift
```diff
@@ -97,3 +109,17 @@ struct ContentSettingsView: View {
     .background(theme.secondaryBackgroundColor)
   }
 }
+
+fileprivate extension View {
+  func onChange(of: some Equatable, _ perform: @escaping () -> Void) -> some View {
+    if #available(iOS 17, macOS 14, *) {
```
Dimillian, line 115: Not shipping yet with Xcode 15. So this can't be used or wrapper yet.

Packages/Env/Sources/Env/UserPreferences.swift
```diff
@@ -25,6 +25,7 @@ public class UserPreferences: ObservableObject {
   @AppStorage("app_auto_expand_spoilers") public var appAutoExpandSpoilers = false
   @AppStorage("app_auto_expand_media") public var appAutoExpandMedia: ServerPreferences.AutoExpandMedia = .hideSensitive
   @AppStorage("app_default_post_visibility") public var appDefaultPostVisibility: Models.Visibility = .pub
+  @AppStorage("app_default_reply_visibility") public var appDefaultReplyVisibility: Models.Visibility = .unlisted
```
Dimillian, line 28: What do you think of this default? I feel like if there was no setting before it should be public. Or at least be initialised the first time with appDefaultPostVisibility

Packages/Env/Sources/Env/UserPreferences.swift
```diff
@@ -154,4 +174,17 @@ public class UserPreferences: ObservableObject {
     copy.insert(isoCode, at: 0)
     recentlyUsedLanguages = Array(copy.prefix(3))
   }
+  
+  public static func getIntOfVisibility(_ vis: Models.Visibility) -> Int {
+    switch vis {
```
Dimillian, line 179: I guess you could try to make Visibility enum an Int instead of String. Not sure we really use its String value for anything at the moment.

Dimillian:
All good thanks!

Packages/Env/Sources/Env/UserPreferences.swift
Havhingstor: The string value seems to be used for decoding from the JSON we get. I’m implementing a manual decoding, I’ll push it shortly, you can decide what you like better

Havhingstor:
I’ve incorporated the third comment too, but it’s probably better and easier to leave it the way it’s merged now👍

PR #1510, opened 2023-07-19

Jerry23011:
### Changes
- Add new translations for Reply Visibility and Tag Groups


nixzhu:
LGTM

PR #1513, opened 2023-07-19
Show all symbols when adding a tag group

Havhingstor:
All the SFSymbols are shown when adding a tag group, the list is taken from SFSafeSymbols. This includes symbols such as "swift" and "apple.logo", as seen in https://mastodon.social/@alexito4/110742407894134083.


Dimillian:
Very nice!

Issue #1546, opened 2023-08-12
Bug: Umlaut in URL not recognized as valid link


AliveDevil:
## Environment:

 - OS: iOS 16.6
 - IceCubesApp version: 1.7.3

## Description

Receive a post containing an umlaut in its path, this URL won’t be recognized as a link.

```
https://en.wikipedia.org/wiki/Elbbrücken_station
```

With URL encoded characters the URL works:
![image](https://github.com/Dimillian/IceCubesApp/assets/734738/b7da4329-d346-4479-bc5e-5733789a0f31)

## Related Issues

 - [x] Search that this bugs don't already exist before creating it.

grant-mcs:
This should now be fixed in the main branch.

PR #1550, opened 2023-08-17
Support status links with non-ASCII characters (Bugfix 1546)

grant-mcs:
Fix for issue opened here: https://github.com/Dimillian/IceCubesApp/issues/1546

This change affects the decoding of statuses as they're received so statuses received and decoded prior to this change being applied will continue to show the previous behaviour.

Prior to fix:

image

After fix:

image


Dimillian:
Thanks for the fix and the tests!

PR #1561, opened 2023-08-23
Bugfix 1459 - Escape ~ character as markdown

grant-mcs:
While it isn't documented in the official markdown spec, the `~` character is often used to denote ~~strikethrough~~ formatting. This includes the AttributedString class. To address this, the `~` is now escaped along with the other markdown characters.

An example of two statuses, one processed before the change and one processed after the change:

image


Dimillian:
Finally, no more Markdown :D

PR #1566, opened 2023-08-30

Jerry23011:
### Changes
- Add Simplified Chinese translation for "Privacy Settings"

@nixzhu if you are available :)


nixzhu:
LGTM

Issue #337, opened 2023-01-24
On iOS 16.2 iPhone Pro 12, when creating a longer post, the screen doesn't scroll when typing hits the bottom. Must manually scroll to see what you are typing.


ddimick:
https://user-images.githubusercontent.com/46494221/214197436-ffae049d-ae47-4771-af87-0171eef8e8a2.MOV

PR #1570, opened 2023-09-11
Fix text field bottom padding in EditorView

eslamNahel:
Even though this might not seem like a "crucial" fix, I took it upon myself to do it anyway. I have some free time these days and decided to do some open source work, and I have been following your journey with building this app for a while now and I am really impressed. So I decided I would jump in to help. 

This is just a first PR that would help me get more familiar with the codebase.
__________

It was mentioned in this issue #337 that the text view doesn't scroll when you are typing and reached the bottom. This was mainly because the ScrollView for the text view has a safe area for the keyboard, but it doesn't account for the extra space for the `StatusEditorAccessoryView`, which made the text fall behind the `StatusEditorAccessoryView` when it reached the bottom.
_________
The straight forward answer would be to add a bottom padding for the scroll view, and it's true. But this resulted in some very weird bug. The scroll view would ignore the navigation bar safe area. See the video 👇🏼

https://github.com/Dimillian/IceCubesApp/assets/54270105/36cacdeb-03ed-4565-8503-22142ca4082f

And apparently, the extra top padding (even for 1) fixes it 👇🏼

https://github.com/Dimillian/IceCubesApp/assets/54270105/5ef7bea3-7a59-45a4-aa5c-3b6e19b537e7

The only down side is that we lose the navigation bar appearance configuration for scrolling.

### DIFF
--- a/Packages/Status/Sources/Status/Editor/StatusEditorView.swift
+++ b/Packages/Status/Sources/Status/Editor/StatusEditorView.swift
@@ -66,7 +66,10 @@ public struct StatusEditorView: View {
           }
           .padding(.top, 8)
           .padding(.bottom, 40)
-        }.accessibilitySortPriority(1) // Ensure that all elements inside the `ScrollView` occur earlier than the accessory views
+        }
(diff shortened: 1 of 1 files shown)


Dimillian:
Thanks!

PR #1593, opened 2023-09-26
Remove "Translate with DeepL"-option

### LINKED ISSUE #1583: Bug: Too Many Translate Buttons
## Environment:

 - OS:  17.0
 
 - IceCubesApp version: 1.8.1
 

## Description

Icecubes is showing twice as many translate buttons as necessary in the per toot … menu. It shows two. The correct behavior is to show exactly one.

Pick an algorithm/service. It doesn’t matter which, but pick one. There is no way for a user to know what the difference between “Translate” and “Translate with DeepL” are, nor why it matters. As it is, it’s a paradox of choice.

Honestly, it reminds me of the old GNOME 1.x menu that featured 7 different clocks, with such useful names as “Clock” and “Another Clock”.

## Related Issues

https://github.com/Dimillian/IceCubesApp/issues/1289

Havhingstor:
The "Translate with DeepL"-option is removed to make the app better understandable for the average user. A person who wants to use DeepL can still insert their own API key to always use DeepL.

Fixes #1583

### DIFF
--- a/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
+++ b/IceCubesApp/Resources/Localization/be.lproj/Localizable.strings
@@ -412,7 +412,6 @@
 
 // MARK: Package: Status
 "status.action.translate" = "Перакласці";
-"status.action.translate-with-deepl" = "Translate with DeepL";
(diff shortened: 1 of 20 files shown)


Dimillian:
Make sense, a win for simplicity!

Issue #1588, opened 2023-09-24
Feature request: Tap on tab to scroll to top (and focus on search field) in all tabs


boscojwho:
1. Currently, tapping on any tab causes that navigation stack to pop to root, but scroll to top is only implemented on Timeline tab.
2. Additionally, tapping on tab to focus on search field (if already scrolled to top) would be nice.

@Dimillian I wouldn't mind implementing this, unless there's a reason why this was omitted in the first place?

Dimillian:
Yes, it's only the home tab because forwarding it to all tabs is a pain. If you find a good way to do it, feel free!

boscojwho:
Ah fair, let me see what the laziest implementation looks like first.
- I was going to just use the same pattern as the home tab, but that would require injecting the same binding multiple times I suppose 🤨

Dimillian:
Well, it's fine to do it this way, it's ugly but we got to work with what we have 😆

boscojwho:
Got it working on all tabs: 
- Need to fix #1592 first before creating a PR for this.

PR #1594, opened 2023-09-27
Update Explore trending links to use value-based NavigationLink

boscojwho:
Fixes trending links "see more" not getting added to navigation path, which means pop to root using tab bar doesn't work here.
- Partially addresses #1588


Dimillian (approved):
Nice, thanks!

PR #1595, opened 2023-09-27
Display the Pending/Unread button at the bottom of the screen

clouetb:
Since I have a big screen and smaller hands, having the Pending/Unread toots at the top of the screen is more comfortable at the bottom of the screen for a single-handed use.

This PR enables toggling such behavior.

I also added translation for the settings screens (using automated translation tools, hence some maybe iffy wording in the languages I don't know).

Please let me know if it would be useful to have the ability to locate the Pending/Unread button in the all four corners of the screen.


Dimillian:
It's all good, thanks!

Issue #1568, opened 2023-09-07
Feature request: Block user confirmation dialog


mdesmedt:
I just accidentally blocked a user after mis-touching "Block" instead of "Message".

Please add a "Are you sure you want to block [user]?" confirmation before blocking a user account.

eric-Is-Me:
I tried my hand at implementing this confirmation dialog locally and it seems to work for the happy path use case. I'd like to open up a PR for this but I don't believe I have permission to create branches on this repository yet. (first time open source contributor here)
@Dimillian is there a process for me to get contributor access so I can push up a branch?

eric-Is-Me:
Created PR: https://github.com/Dimillian/IceCubesApp/pull/1597

eric-Is-Me:
PR that implements this feature request has been merged to mainline.
https://github.com/Dimillian/IceCubesApp/commit/4bbfdcd256ca4271b7dd5a53bf0b582e8131b017

@mdesmedt lets close this issue out?

mdesmedt:
Cool. Thanks!

PR #1597, opened 2023-09-28
Feature request: Block user confirmation dialog

eric-Is-Me:
- Creates and uses a new StateObject called BlockConfirmation which holds a value to signal when to show the confirmation dialog.
    - Moved the post request for blocking out of the ContextMenu since the Menu view within AccountDetailView will close as soon as one of it's items is clicked.

Note: The code I wrote does not specify thread priority or anything so I think the toggle of the `isPresentingBlockingConfirmation` bool could be done concurrently which may lead to unexpected behavior. Not sure if we want to mark the `BlockConfirmation` class as a MainActor for this.

This PR is in response to issue #1568

### DIFF
--- a/Packages/Account/Sources/Account/AccountDetailView.swift
+++ b/Packages/Account/Sources/Account/AccountDetailView.swift
@@ -20,8 +20,7 @@ public struct AccountDetailView: View {
   @Environment(Client.self) private var client
   @Environment(RouterPath.self) private var routerPath
 
-  @StateObject private var blockConfirmation: BlockConfirmation = BlockConfirmation()
-  
+  @State private var showBlockConfirmation: Bool = false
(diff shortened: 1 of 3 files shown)


Packages/Account/Sources/Account/AccountDetailView.swift
```diff
@@ -400,6 +418,10 @@ extension View {
   }
 }
 
+class BlockConfirmation: ObservableObject {
+  @Published var isPresentingBlockingConfirmation: Bool = false
+}
```
Dimillian, line 423: This should be directly in the view as a simple State property and passed to the context menu as a binding. 
`ObservableObject` is also not used anymore as the project was migrated to the new observation framework.

PR #1598, opened 2023-09-29
Feature: Tab bar scroll to top

boscojwho:
## About this Pull Request
What's New:
- Users can now tap on any tab to "pop to root", then scroll-to-top.
- Additionally, in `Explore` tab, once scrolled-to-top, tapping on that tab again causes the search field to become focussed.

## Additional Notes
Unfortunately, in `Explore` tab, when the search interface is presented (but isn't "first responder"), there is no way to programmatically re-focus the search bar.
- Also,  in `Explore`'s search results list, tapping on that tab currently does nothing. I wasn't sure if it made sense there, since it would set up the expectation that tapping on that tab again would re-focus on the search bar (which isn't doable with the current APIs).

### DIFF
--- a/Packages/Notifications/Sources/Notifications/NotificationsListView.swift
+++ b/Packages/Notifications/Sources/Notifications/NotificationsListView.swift
@@ -212,10 +212,7 @@ public struct NotificationsListView: View {
   }
   
   private var scrollToTopView: some View {
-    HStack { EmptyView() }
-      .listRowBackground(theme.primaryBackgroundColor)
-      .listRowSeparator(.hidden)
-      .listRowInsets(.init())
+    ScrollToView()
       .frame(height: .scrollToViewHeight)
       .id(Constants.scrollToTop)
       .onAppear {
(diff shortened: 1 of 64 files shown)


Packages/Notifications/Sources/Notifications/NotificationsListView.swift
```diff
@@ -196,4 +210,16 @@ public struct NotificationsListView: View {
       .frame(height: .layoutPadding)
       .accessibilityHidden(true)
   }
+  
+  private var scrollToTopView: some View {
+    ScrollToView()
+      .frame(height: .scrollToViewHeight)
+      .id(Constants.scrollToTop)
```
Dimillian, line 217: So, the id could be directly set in the `ScrollToView` you created. You should also move the enum here so you don't duplicate it.

boscojwho: Yea, that's an easy change.

PR #1601, opened 2023-10-01
Fix a crash bug at `Client.makeURL`

yasuradodo:
## Issue

The crash will happen when you type something unexpected instance URL.

Screenshot 2023-10-01 at 22 02 26

Example
```swift
let server = "test.com/"

var components = URLComponents()
components.scheme = "https"
components.host = server
components.path = "/api/v1/instance"
components.url! // 💥 error: Execution was interrupted, reason: EXC_BREAKPOINT (code=1, subcode=0x18c986650).
```

## Solution
Throwing an error instead of doing force unwrap.

### DIFF
--- a/Packages/Env/Sources/Env/StreamWatcher.swift
+++ b/Packages/Env/Sources/Env/StreamWatcher.swift
@@ -40,9 +40,14 @@ import Observation
   }
 
   private func connect() {
-    guard let client else { return }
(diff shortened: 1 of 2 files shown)


lgreydev:
@yasuradodo Thank you buddy!

Dimillian:
This is good, thanks!

PR #1611, opened 2023-10-05
de: Update Localizable.xcstrings

dertuxmalwieder:
Added tag group translations.

*NOTE:* The switch to a string catalog has a few significant disadvantages: 1) no granularity (all translators must submit a complete localization file - enjoy merge conflicts :p), 2) it has become very hard to maintain a translation in a simple text editor (Xcode is probably the least annoying option now).


Dimillian:
Yep it's probably a bit harder to manage without Xcode. But it's so much better using it it's worth it. Merge conflict should be fine due to the structure of the file. Guess will see.

PR #1612, opened 2023-10-05

Jerry23011:
### Changes
- Complete Simplified Chinese localization strings
- Add translation for `Add or remove tag from tag groups`

@nixzhu String Catalogs! 😁


nixzhu:
LGTM

PR #1617, opened 2023-10-10
Update Simplified Chinese localization for block user confirmation

Jerry23011:
### Changes
- Add SC translations for block user confirmation

@nixzhu Could you review this if you're free :)


nixzhu:
LGTM

PR #1626, opened 2023-10-17
New App Icon Gallery

Bonney:
Was hacking around with a UI for picking app icons over the weekend, and it felt like a great fit for IceCubes, with it's selection of community icons.

All relevant code lives in the `AppIcons` package and is completely self-contained with no dependencies.

Short demo screen recording showing the new interface: https://github.com/Dimillian/IceCubesApp/assets/3890339/1cc8c1e5-a463-4659-98a9-00d27334ff28


Dimillian:
I like the UI, but I think there is a few shortcomings that would be nice to improve on: 
* The number of rows should be dynamic. Maybe there is something to do with ViewThatFit or Grid. It would be better looking on iPad / macOS if it was dynamic. 
* The size of the icons in the grid should also be dynamic. Maybe 2/3 sizes would suffice, but I feel like it's a bit small on iPad. 

Appart from that it's truly an improvement over the current selector.

Bonney:
Good feedback, I’ll make some tweaks for those larger sizes. :) – Matt BonneyOn Oct 18, 2023, at 2:57 AM, Thomas Ricouard ***@***.***> wrote:﻿
I like the UI, but I think there is a few shortcomings that would be nice to improve on:

The number of rows should be dynamic. Maybe there is something to do with ViewThatFit or Grid. It would be better looking on iPad / macOS if it was dynamic.
The size of the icons in the grid should also be dynamic. Maybe 2/3 sizes would suffice, but I feel like it's a bit small on iPad.

Appart from that it's truly an improvement over the current selector.

PR #1644, opened 2023-10-26
Fix flickering issues when resizing window

HugoSay:
On macOS and probably iPadOS.

As shown in the video, the processor doesn't appear to change the image rendering but it does create an annoying flicker 😱

If it's really useful for performance, I guess I can improve the fix but it seemed fine to just remove it. 

Before: 

https://github.com/Dimillian/IceCubesApp/assets/10938182/4bf0a8b5-b729-4796-83a6-251b3bbea7de

After:

https://github.com/Dimillian/IceCubesApp/assets/10938182/0c12b9df-b481-41ac-8b48-c7607666e185

### DIFF
--- a/Packages/DesignSystem/Sources/DesignSystem/Views/LazyResizableImage.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/Views/LazyResizableImage.swift
@@ -1,0 +1,42 @@
+//
+//  LazyResizableImage.swift
+//
+//
+//  Created by Hugo Saynac on 28/10/2023.
+//
+
+import Nuke
+import NukeUI
+import SwiftUI
+
+/// A LazyImage (Nuke) with a geometry reader under the hood in order to use a Resize Processor to optimize performances on lists.
+/// This views also allows smooth resizing of the images by debouncing the update of the ImageProcessor. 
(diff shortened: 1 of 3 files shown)


Dimillian:
Let me ping @kean to see what he think.

Dimillian:
FYI @HugoSay, It's really needed for performances, especially on older devices; it greatly improves scroll performances and memory management. I think a solution would be to debounce the processors, so the size is not updated at every window frame update. I'll look into it.

kean:
I haven't considered resizing when adding these resizing processors. The best solution would be to specify the preferred target size based on the size breakpoints and have only a small preset of the preferred sizes. If it's an iPhone, it could continue working as is as it has no resizing.

kean:
> I think a solution would be to debounce the processors, so the size is not updated at every window frame update.

Debouncing should also work and it may be a better solution overall.

Removing the resizing entirely is not advised.

HugoSay:
I just updated the PR, adding some debouncing logic in a `LazyResizableImage` view.
(the struct needs to be moved somewhere else, but I wanted to get your feedback on that first 🙂)

Dimillian:
Looking at this soon, thanks!

Dimillian:
Thanks! It's much better!

PR #1646, opened 2023-10-26

Jerry23011:
### Changes
- Add translation for `New Window` on macOS, refered to Safari's translation

@nixzhu If you have time :)


nixzhu:
LGTM

PR #1651, opened 2023-11-01
Refactor MediaUIView state and logic

thai-d-v:
I refactored `MediaUIView` to make it more logical and maintainable.

1. use the new data type `DisplayData` because:
    - `MediaUIView` always needs a `url` and a valid `supportedType` to display an attachment.
    - Its job is displaying, not checking. `DisplayData` filters out invalid data for it.

2. `isSavingPhoto` and `didSavePhoto` actually reflect only one state, so I use a new enum for it
```swift
  private enum SavingState {
    case unsaved
    case saving
    case saved
  }
```

3. use the item directly instead of its id in `scrollToId` (now is `scrolledItem`) to avoid reaching on toolbar code like this:
```swift
if let alt = attachments.first(where: { $0.id == scrollToId})?.description {...}
```

4. extract subview to
    - narrow down scope states, each state should only exist in the right and narrow scope
    - make code structure more clear

### DIFF
--- a/Packages/MediaUI/Sources/MediaUI/MediaUIView.swift
+++ b/Packages/MediaUI/Sources/MediaUI/MediaUIView.swift
@@ -1,78 +1,64 @@
-import Foundation
-import NukeUI
 import Nuke
 import SwiftUI
 import Models
 import QuickLook
 
 public struct MediaUIView: View, @unchecked Sendable {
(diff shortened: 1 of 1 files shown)


Dimillian:
It's solid! I like it. Thanks. Testing a few things and I'll merge.

PR #1657, opened 2023-11-04
I removed "Done" button in setting tab on iPhone

thai-d-v:
I removed the "Done" button because it has no effect since "Setting" is a separate tab (on iPhone). The reason is that the `@Environment(\.dismiss)` action is only meant to be used for:
- Dismiss a modal presentation, like a sheet or a popover.
- Pop the current view from a [NavigationStack](https://developer.apple.com/documentation/swiftui/navigationstack).
- Close a window that you create with [WindowGroup](https://developer.apple.com/documentation/swiftui/windowgroup) or [Window](https://developer.apple.com/documentation/swiftui/window).

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
+++ b/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
@@ -12,7 +12,6 @@ import Timeline
 
 @MainActor
 struct SettingsTabs: View {
-  @Environment(\.dismiss) private var dismiss
   @Environment(\.modelContext) private var context
 
   @Environment(PushNotificationsService.self) private var pushNotifications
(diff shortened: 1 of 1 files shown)


Dimillian:
I am closing this one. This is a standard UX on iOS. Even if you can pull down to close the sheet Apple recommends clear button to close it. 
On iPhone the setting is not a tab but presented as a sheet.

thai-d-v:
should I make it "go back to the main tab" when users tab it?

thai-d-v:
Sorry I misspelled it, I meant “tap”. I think that if you want to keep the “Done” button, it is better to do the action “going back to the main tab” when users tap/touch it. What do you think about it?

PR #1662, opened 2023-11-07
Improve SoundEffectManager & HapticManager

tarrouye:
1. Refactor the way `SoundEffectManager` registers and plays sounds
- This change makes it much easier to add a new sound
- Before: Had to add a new case to the enum, add a new var for the soundID, and add the new case to the switch in register and play functions. Now you can `.playSound` the new sound effect
- After: Simply add a new case to the enum. Nothing else is required, now you can `.playSound` it freely !

2. Adjust the call-site
- IMO `playSound(of: SoundEffect)` is odd, `of` doesn't feel like the appropriate prefix. I would say `playSound(for: SoundEffect)` makes for sense, but I also prefer simply `playSound(SoundEffect)` , since the type is kind of explicit and the prefix feels unnecessary. 
- Unlike change 1, this is completely subjective and has no tangible benefits. I'm happy to revert this part if you prefer the previous API. 

3. Adjust `HapticManager` call-site to match change made to `SoundEffectManager`
- Same reasoning, same caveat - happy to revert this part. 

I split this into 3 commits matching the 3 points above; The first commit contains the main useful change, whereas the second and third are entirely stylistic and I'd be happy to revert depending on your feedback.

### DIFF
--- a/IceCubesApp/App/Main/IceCubesApp+Tabbar.swift
+++ b/IceCubesApp/App/Main/IceCubesApp+Tabbar.swift
@@ -14,7 +14,7 @@ extension IceCubesApp {
         }
       }
 
-      HapticManager.shared.fireHaptic(of: .tabSelection)
+      HapticManager.shared.fireHaptic(.tabSelection)
       SoundEffectManager.shared.playSound(.tabSelection)
 
       selectedTab = newTab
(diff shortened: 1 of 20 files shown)


Dimillian:
It's the refactor where I think, "Why did I not do that in the first place." I guess the CFURL threw me off. Really nice, thanks!

Issue #1627, opened 2023-10-18
Feature request: Better notification filters


tehabe:
Currently the filters for the Notifications are resetting to "all" after you leave the app. That is very annoying. I only want to see mentions and I only want to be notified when someone mentions me. But there is currently no way of doing that.

PR #1663, opened 2023-11-07
Feature: store selected notification filter (#1627)

thai-d-v:
I implement the feature request in #1627. I follow the convention in `TimelineCache` that uses `UserDefaults.standard` to store data.


thai-d-v:
Just in case you missed this, does this PR go in the right direction @Dimillian?

Dimillian:
Oh sorry, I forgot to add a comment on this one. I think I'm fine with it but maybe a question: Are you sure you want to save the setting per account? I think that when you switch account it should keep it on the same filter.

thai-d-v:
Ah, you made a good point! I updated it.

Issue #1554, opened 2023-08-19
Feature request: Timeline media should be limited to screen height (iPad)


sh95014:
On a landscape-orientation iPad in particular, portrait images or videos can be as much as two screen heights tall, and often displayed in poor upsampled resolution. It should be a much better general experience to limit the height of images and videos to the screen height.

In the screenshot below, for example, you see barely half of the video (but plenty of artifacts!):

![IMG_2D52D3281176-1](https://github.com/Dimillian/IceCubesApp/assets/95387068/55fc3b8b-81c9-4b64-be81-09434785ebd1)

sh95014:
Height computation is gross, but I think this is already better:

```diff
diff --git a/Packages/DesignSystem/Sources/DesignSystem/SceneDelegate.swift b/Packages/DesignSystem/Sources/DesignSystem/SceneDelegate.swift
index 44881c96..c93ba21d 100644
--- a/Packages/DesignSystem/Sources/DesignSystem/SceneDelegate.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/SceneDelegate.swift
@@ -8,6 +8,10 @@ import UIKit
     window?.bounds.size.width ?? UIScreen.main.bounds.size.width
   }
 
+  public var windowHeight: CGFloat {
+    window?.bounds.size.height ?? UIScreen.main.bounds.size.height
+  }
+
   public func scene(_ scene: UIScene,
                     willConnectTo _: UISceneSession,
                     options _: UIScene.ConnectionOptions)
diff --git a/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift b/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
index d4a04b50..c2721550 100644
--- a/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
+++ b/Packages/Status/Sources/Status/Row/Subviews/StatusRowMediaPreviewView.swift
@@ -68,7 +68,8 @@ public struct StatusRowMediaPreviewView: View {
           imageMaxHeight: imageMaxHeight,
           sensitive: sensitive,
           appLayoutWidth: appLayoutWidth,
-          availableWidth: availableWidth
+          availableWidth: availableWidth,
+          availableHeight: sceneDelegate.windowHeight - 200
         )
         .accessibilityElement(children: .ignore)
[...]

PR #1675, opened 2023-11-16
Limit image height to screen height

sh95014:
- limit available height to 80% of screen/window height
- if image fits in available width and height, just display it at 1x (to avoid ugly resizing artifacts)
- otherwise, shrink it proportionally to fit

Fixes [#1554](https://github.com/Dimillian/IceCubesApp/issues/1554)


Dimillian:
I love it. Thanks!

PR #1680, opened 2023-11-19
Categorized custom emoji list

deholic:
Users often had a hard time finding specific Mastodon custom emojis because there wasn't a way to organize them. This update adds categories, making it way easier to discover and use the emojis you're looking for.

See It In Action:

https://github.com/Dimillian/IceCubesApp/assets/712513/e3da4759-d777-4356-99d7-089d4fcd44ab


Dimillian:
Thanks! A lot of people were asking for it!

PR #1684, opened 2023-11-21

Jerry23011:
### Changes
- Add new translations


nixzhu:
LGTM

PR #1686, opened 2023-11-23
Fix `EditTagGroupView`

thai-d-v:
### Description

1. refactor code
2. make code safe by:
    - adding guard conditions
    - formating data before saving
    - remove duplicate data
    - fix some bugs, and details in commits' messages
3. make UI easier to understand by:
    - adding warning labels to show what error is
    - moving the symbol search results to the right below the text field to make the layout more coherent
    - fix the focusing flow to make the form easy to navigate by keyboard 

### Demonstration

> You should wait a little bit for the GIF to be loaded.

Kapture 2023-11-23 at 22 00 23


Dimillian:
It's nice, thanks! There is a bit of work for localizations but it can be done later.

Issue #1459, opened 2023-07-02
Feature request: Support for formatted post display


imdatceleste:
Mastodon stores the toots in HTML if there is any format information (e.g. markdown from the `glitch-soc` fork). 

Currently IceCubesApp only displays these toots in plain-text format, all formatting from the server are discarded.

It would be awesome, if it would display the posts formatted with HTML.

AFAIK, when posting a toot, it is sent in plaintext and depending on the server-configuration, the server then converts it to HTML if it detects any formatting (and the user has set it so on their server-profile in e.g. `glitch-soc`).

Thank you for your consideration.

ShadSterling:
Apparently there is some support for formatting; I'm seeing crossed out text in a port that doesn't have any formatting - https://mastodon.social/@glyph/110724377873612649

imdatsolak:
> Apparently there is some support for formatting; I'm seeing crossed out text in a port that doesn't have any formatting - https://mastodon.social/@glyph/110724377873612649

This is a bug, IceCubesApp should **not** interpret markdown by itself. 

If the user has set that they are posting Markdown or HTML (it is a setting in `glitch-soc` at the very least), Mastodon automatically parses the incoming plain-text as being in that format and converts it to HTML before storing it.

So, if IceCubesApp retrieves a toot-text from Mastodon-server, it should not do any additional parsing/interpreting. It should just support the display of HTML in the post-view.

ShadSterling:
How can I distinguish whether the crossed-out text is something Ice Cubes added by interpreting the plain text or something Ice Cubes displayed from the HTML?

imdatsolak:
> How can I distinguish whether the crossed-out text is something Ice Cubes added by interpreting the plain text or something Ice Cubes displayed from the HTML?

If you see it in IceCubesApp, then it means it is how it should be shown, no further interpretation.

Interpretation and conversion from Markdown to HTML is only done **on the server**. If the server returns a ˜text˜ then the tilde-symbols are to be shown and not interpreted in the client (note: I am using the "special" ˜ to avoid GitHub interpreting it).

So, rule:
1. User types text with markdown. Client sends it to the server as PLAIN-TEXT format. Client does **not** interpret anything, does **not** convert anything. It sends it "raw" to the server.
2. Server checks what settings the user has and depending on the settings, it either converts Markdown to HTML **or** leaves it as is.
[...]

PR #1690, opened 2023-11-27
Improve display of HTML ul (bullet list) and ol (numbered list)

nathreed:
While SwiftUI's `Text` view won't display these in an `AttributedString` even if they get parsed from Markdown (which would also require the use of the `.full` option instead of the `.inlineOnlyPreservingWhitespace` option), we can improve the appearance somewhat. 
Currently, list elements are clumped together with no spaces between them, and there's no indication whatsoever that the author indicated these to be a list. 
Change to insert Markdown list syntax with linebreaks and dashes, so users can at least understand there's a list there. Similar change for ordered lists.

This will still be broken for nested lists, but it didn't seem worth it to put a lot of effort into this (or other revamps, like making bold/italics/code work properly) because it seems like the current text handling in Ice Cubes is suboptimal and eventually slated for improvement (according to https://github.com/Dimillian/IceCubesApp/issues/1459#issuecomment-1638562657). 
So this is more designed to make lists "less broken" in some (many) cases, rather than be a comprehensive fix for all lists in all cases.


Dimillian:
Intelligent and good thanks! 
I wonder when I'll have time to rethink the text rendering tho haha.

Issue #1668, opened 2023-11-12
Bug: macOS images dont resize correctly 


markbattistella:
## Environment:

 - OS: macOS Sonoma 14.0
 - IceCubesApp version: 1.9.6

## Description

If a posted image is larger than the width of the app, then it doesnt scale down but instead pushes it off the screen. In a collection of images though, they do scale, but it is the single images which do not.

## Related Issues

 - [x] Search that this bugs don't already exist before creating it.
 

https://github.com/Dimillian/IceCubesApp/assets/17962160/fc06445d-4832-4047-8c8d-f652074455d2

Issue #1656, opened 2023-11-03
Bug: some posts are cut off when rotating from landscape to portrait


andrewbaker-uk:
## Environment:

 - OS: iPadOS 17.1
 - IceCubesApp version: 1.9.6

## Description

When rotating the iPad Pro from landscape to portrait some posts are cut off, because the side pane is on top of the post. This only happens to some posts.

![image](https://github.com/Dimillian/IceCubesApp/assets/10560135/c5884405-c5b3-4c28-9348-30fb80c84874)

## Related Issues

 - [X] Search that this bugs don't already exist before creating it.

andrewbaker-uk:
Closing the app and reopening it does resolve the issue. But there may be a issue

Dimillian:
Yep known issue with the image. You can also scroll them outside of the view and scroll back to them it should fix it.

Issue #1401, opened 2023-04-28
Bug: Width is causing content to be cropped of the leading edge


brennanMKE:
## Environment:

 - OS: macOS
 
 - IceCubesApp version: 1.6.15 (1891)
 

## Description

When I resize the window it does not fit the posts in the new width.

<img width="550" alt="Ice Cubes Layout" src="https://user-images.githubusercontent.com/277419/235206619-2d6e4a58-037e-4d5d-8889-c8045ac5437f.png">

## Related Issues

 - [ ] Search that this bugs don't already exist before creating it.

Dimillian:
Will be fixed in newt version

PR #1693, opened 2023-11-28
Fix: media views don't resize when the window size (device orientation) changes

thai-d-v:
### Description

This PR fixes the media resizing bug for #1401 #1656  #1668.

The reason is because of `SceneDelegate`. Its properties `windowWidth` and `windowHeight` are not observable. They don't emit any change to observers(views).

So I fixed `SceneDelegate` only, I don't change any logic of the view. Any bug related to the view logic, if exists, should be fixed in other PRs.

#### macOS

> You should wait a little bit for the GIF to be loaded.

Resize-macOS--2023-11-28 at 22 13 30

#### iPadOS

> You should wait a little bit for the GIF to be loaded.

Resize-iPadOS--2023-11-28 at 22 40 08

### DIFF
--- a/Packages/DesignSystem/Sources/DesignSystem/SceneDelegate.swift
+++ b/Packages/DesignSystem/Sources/DesignSystem/SceneDelegate.swift
@@ -1,16 +1,11 @@
 import Combine
 import UIKit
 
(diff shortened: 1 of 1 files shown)


Dimillian:
Looking at this one soon, I need to do some performances test.

PR #1696, opened 2023-11-29

Jerry23011:
### Changes
- Add new translations for List settings
- Add new translations for Tab Groups setting

@nixzhu If you have time :)

@Dimillian Is it possible that the new strings in string catalog are not labeled as "translated" when they get synced? It would make the workflow much more convenient with a "NEW" labeled on the untranslated strings.

### DIFF
--- a/IceCubesApp/Resources/Localization/Localizable.xcstrings
+++ b/IceCubesApp/Resources/Localization/Localizable.xcstrings
@@ -14,6 +14,12 @@
             "state" : "translated",
             "value" : ""
           }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : ""
+          }
         }
       }
     },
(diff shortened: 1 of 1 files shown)


IceCubesApp/Resources/Localization/Localizable.xcstrings
```diff
@@ -348,7 +354,7 @@
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "%@ is already selected"
+            "value" : "已选择：%@"
```
nixzhu, line 357: 这个句子是否用于提示？可能`已选过 %@`更好

Jerry23011: 在这里，我觉得这个意思是选择了哪个 Symbol？
SCR-20231128-rnso

nixzhu:
LGTM

Dimillian:
@Jerry23011 I'll try, but the problem is that if I leave them empty then if I release the app it shows the key instead of the default English :/ I don't know why. 
What I'll do is mark them as needing review so at least you can sort them. Sorry about that!

Jerry23011:
> but the problem is that if I leave them empty then if I release the app it shows the key instead of the default English :/

So that's why!

> What I'll do is mark them as needing review so at least you can sort them.

That would be great! Thanks!

PR #1702, opened 2023-12-03
Fix the lines of indentation

thai-d-v:
I connect the lines of indentation to create a seamless flow.

Simulator Screenshot - iPhone 15 Pro Max - 2023-12-03 at 16 02 34


Dimillian:
It would be nice, and I've tried it before, but it breaks too many things. In the timeline and notifications now the custom divider is not edge-to-edge anymore. Thez spacing between rows is also too close. 

image

image


And in notifications it's weird as the row is highlighted when it's a mention.

thai-d-v:
Ok, I close this one. Maybe We'll come up with a better solution someday. 😁

PR #1703, opened 2023-12-03
Using BUNDLE_ID_PREFIX to fix fork building with iCloud container

Dean151:
Currently, the instructions to build "out of the box" won't work anymore.
Using the prefix seems to fix it.


Dimillian:
Hmmm but that will change iCloud container for everyone right?

Dean151:
Haven't noticed that it was both release & debug. Was persuaded it wasn't. Don't merge as is. Going to change other things to comply with debug vs release

PR #1704, opened 2023-12-03
Added a "Continues a thread" label to status rows

Dean151:
Just like "boosted by" and "replied to", I do think that thread label brings value when seeing a message that might be out of context in the timeline. This helps bringing context without having to press the message to check if there is a thread above. Pressing the message will do the same and allow to consult the whole thread at a glance.

IMG_D95FC98160A5-1

### DIFF
--- a/IceCubesApp/Resources/Localization/Localizable.xcstrings
+++ b/IceCubesApp/Resources/Localization/Localizable.xcstrings
@@ -65752,9 +65752,45 @@
     "status.row.is-thread" : {
       "extractionState" : "manual",
       "localizations" : {
+        "be" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "Continues a thread"
+          }
+        },
+        "ca" : {
+          "stringUnit" : {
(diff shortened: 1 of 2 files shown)


Dimillian:
I like it. But to not we already had the thread icon near the ...

Dean151:
> I like it. But to not we already had the thread icon near the ...

Never noticed it ; and also not distinguishing replies & threads.

Now I know that "reply label missing + discussion icon" means thread … but somehow it feels wrong with me ^^

IceCubesApp/Resources/Localization/Localizable.xcstrings
```diff
@@ -65749,6 +65749,23 @@
...
+          }
+        },
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Complète un fil"
+          }
+        }
+      }
+    },
```
Dimillian, line 65768: Unfortunately you need to add all translation with the default English. If not it'll show the label key :/

PR #1722, opened 2023-12-07
Improve media selection on the status editor.

thai-d-v:
### Description

This change doesn't affect the existing media menu.

There are issues with the media selection in the status editor, which negatively impacts the app's user experience and may hinder its success in the current AI trend.

#### Performance and logical problems

- lagging UI when showing media after selecting because preparing tasks (creating containers, uploading media) don't run in parallel
- not removing photo pickers (not deselecting on the photo app) when removing media on the post editor
- pickers don't have identifiers after being selected
- re-preparing(creating containers, uploading media) the whole media list every time adding new ones

#### UI problems

image

- lack of a menu button (delete, alt) directly on the media item may be counter-intuitive in comparison to the current trend of UI design
- the alt button only appears when the alt-text is not empty, but it does nothing, confusing users. It now always shows up and will lead users to the "Edit Media" view

### Miscellaneous

- I leave the measurement code for creating containers at 2bc49fbe54dd37b1009380f28fa4f054db3dc13c. It is now at least 6 times faster.

### DIFF
--- a/Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
+++ b/Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
@@ -563,20 +563,7 @@ import SwiftUI
 
   func prepareToPost(for pickerItem: PhotosPickerItem) {
     Task(priority: .high) {
-      var container: StatusEditorMediaContainer?
-
-      let clock = ContinuousClock()
-      var elapsed = await clock.measure {
-        container = await old_makeMediaContainer(from: pickerItem)
-      }
-      print("old implementation: \(elapsed)")
-
-      elapsed = await clock.measure {
-        container = await makeMediaContainer(from: pickerItem)
-      }
-      print("new implementation: \(elapsed)")
-
-      if let container {
+      if let container = await makeMediaContainer(from: pickerItem) {
         await MainActor.run { self.mediasImages.append(container) }
         await upload(container: container)
         await MainActor.run { self.isMediasLoading = false }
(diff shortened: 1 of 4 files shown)


Dimillian:
I'm excited to look into this. One of the worst part of the app. Code I wrote so fast to go to market I wanted to forget about it. Props at making it better!

Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
```diff
@@ -84,19 +84,30 @@ import SwiftUI
...
-  var selectedMedias: [PhotosPickerItem] = [] {
+  var mediaPickers: [PhotosPickerItem] = [] {
     didSet {
-      if selectedMedias.count > 4 {
-        selectedMedias = selectedMedias.prefix(4).map { $0 }
+      if mediaPickers.count > 4 {
+        mediaPickers = mediaPickers.prefix(4).map { $0 }
+      }
+
+      let removedIDs = oldValue
```
Dimillian, line 93: Nice yeah. The fact that it was staying selected in the system picker was really bugging me off!

Packages/Status/Sources/Status/Editor/StatusEditorViewModel.swift
```diff
@@ -532,93 +557,111 @@ import SwiftUI
...
-    mediasImages.firstIndex(where: { $0.id == container.id })
+    mediaContainers.firstIndex(where: { $0.id == container.id })
   }
 
-  func inflateSelectedMedias() {
-    mediasImages = []
+  func prepareToPost(for pickerItem: PhotosPickerItem) {
+    Task(priority: .high) {
+      if let container = await makeMediaContainer(from: pickerItem) {
+        await MainActor.run { self.mediaContainers.append(container) }
```
Dimillian, line 566: Is it necessary? `StatusEditorViewModel` being `MainActor` should be enough for the property mutation to happen on the main thread.

Dimillian (approved):
Really awesome. I like the code, and the new inline delete and alt is super cool! Thanks!

PR #1728, opened 2023-12-08
Relayout media on status editor

thai-d-v:
- re-layout the media to optimize the content display
- animate the layout to improve user experience

relayout media 2

### DIFF
--- a/Packages/Status/Sources/Status/Editor/Components/StatusEditorMediaView.swift
+++ b/Packages/Status/Sources/Status/Editor/Components/StatusEditorMediaView.swift
@@ -15,31 +15,51 @@ struct StatusEditorMediaView: View {
   @State private var isErrorDisplayed: Bool = false
 
   var body: some View {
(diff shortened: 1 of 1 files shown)


thai-d-v:
I think scrolling for just 4 items is little bit overkill. I just re-layout, I don't change the items’ behavior.

Dimillian:
It was more like scrolling as soon as it's not one item

thai-d-v:
I understand your point now. It would be better to make the UI more consistent.

- I simplified the resizing process compared to the timeline media to avoid distractions.
- I added a scroll indicator for Mac users because they need it.
- The view now auto-scrolls to newly added media to help users confirm that the media has been added.
- I also added animations to help users understand what's happening.

https://github.com/Dimillian/IceCubesApp/assets/46838577/307bcd21-017e-43ad-adbd-0c3b9b1beb9b

Dimillian:
Thanks!

PR #1731, opened 2023-12-10
Add Select Status Text Action

thai-d-v:
I just added the select text action, existing actions are unchanged.

- follow the current theme convention
- support attributed string to give the same text appearance as status
- already localized (English)
- tested on both iOS and macOS

(video)

https://github.com/Dimillian/IceCubesApp/assets/46838577/d888d4ac-3173-4e74-8255-42412f37ec5d


Dimillian:
* Fixed dark mode
* Fixed localization

Thanks! This is exactly how I was thinking to do it haha!

PR #1736, opened 2023-12-13
Link to parent post

Havhingstor:
The "reply to ..."-text is now a link to the parent post. A tap scrolls to the parent if the whole hierarchy over a post is shown (detail view). Otherwise, the detail view for the parent is opened.


Dimillian:
This is really nice!

PR #1737, opened 2023-12-13
Allow the user to customize the thread indentation

Havhingstor:
The user can now select if they want to indent threads/replies, and how much
the replies should be indented.


Dimillian:
perfect!

Issue #1735, opened 2023-12-12
Feature request: Multi Toots (Posts)


detrophy:
Please add the possibility to post multiple posts directly as a thread to allow for users on standard character count instances to easily create threads. 

Maybe don't ask the user to create toots by, but create additional ones as needed.

Add either 🧵 or N/N as thread indicator at the end of each post.

![image](https://github.com/Dimillian/IceCubesApp/assets/23260830/44804f74-e220-4199-a9f3-c5e54114358e)

Dimillian:
This is planned :)

PR #1740, opened 2023-12-14
Feature: Post and Reply by Threads

thai-d-v:
@Dimillian Sorry, I stated this PR before #1735 opened, and I just read it after I submitted this PR. I noticed that you have already self-assigned. In case my work does not align with your plan, I will close it.

This feature allows users to post and even reply to a status by a thread.
- all statuses are fully functional individually
- all statuses have the same visibility
- the maximum number of follow-up statuses is 5, be cautious when increasing it
- already localized

https://github.com/Dimillian/IceCubesApp/assets/46838577/def71c4b-c941-4b32-b7b8-a7550c73497b

### DIFF
--- a/Packages/Status/Sources/Status/Editor/StatusEditorView.swift
+++ b/Packages/Status/Sources/Status/Editor/StatusEditorView.swift
@@ -15,52 +15,118 @@ import UIKit
 public struct StatusEditorView: View {
   @Environment(AppAccountsManager.self) private var appAccounts
   @Environment(CurrentAccount.self) private var currentAccount
+  @Environment(Theme.self) private var theme
 
-  @State private var viewModel: StatusEditorViewModel
-  @FocusState private var isSpoilerTextFocused: Bool
-  @State private var editingContainer: StatusEditorMediaContainer?
+  @State private var mainSEVM: StatusEditorViewModel
+  @State private var followUpSEVMs: [StatusEditorViewModel] = []
+  @FocusState private var isSpoilerTextFocused: UUID? // connect CoreEditor and StatusEditorAccessoryView
+  @State private var editingMediaContainer: StatusEditorMediaContainer?
+  @State private var scrollID: UUID?
+
+  @FocusState private var editorFocusState: EditorFocusState?
+  private var focusedSEVM: StatusEditorViewModel {
+    if case let .followUp(id) = editorFocusState,
+       let sevm = followUpSEVMs.first(where: { $0.id == id })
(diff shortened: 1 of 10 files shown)


Dimillian:
Impressive!

Packages/Status/Sources/Status/Editor/StatusEditorView.swift
```diff
@@ -14,230 +14,350 @@ import UIKit
...
+  @State private var mainSEVM: StatusEditorViewModel
+  @State private var followUpSEVMs: [StatusEditorViewModel] = []
+  @FocusState private var isSpoilerTextFocused: UUID? // connect CoreEditor and StatusEditorAccessoryView
+  @State private var editingMediaContainer: StatusEditorMediaContainer?
+  @State private var scrollID: UUID?
 
-  @State private var isDismissAlertPresented: Bool = false
-  @State private var isLanguageConfirmPresented = false
+  @FocusState private var editorFocusState: EditorFocusState?
+  private var focusedSEVM: StatusEditorViewModel {
```
Dimillian, line 27: This is really nice!

thai-d-v:
@Dimillian I am facing issues with localized strings. Normally, when I add a new code with `LocalizedStringKey`, Xcode automatically adds that key to the string catalog file. However, lately, Xcode is not doing it for me. Do you have any idea why?

Dimillian:
Same and no, all the settings are on so I don't know why. 
Also usually if there is a missing translation iOS is supposed to use the en base, but it's not, so I need to add the default English to every language.

Dimillian:
Just renaming a few stuff and moving around files. I think it's good! 
I'm wondering if we really need a main VM and the array, could we just have an array of ViewModel with 0 being the main?

thai-d-v:
I intend to restructure the UI to make it easier for me to understand the code when I come back to it later. This will also make the logic code safer and easier to write. For instance, with this new code, it will be less likely to accidentally remove the main core editor.

https://github.com/thai-d-v/IceCubesApp/blob/4efe32c53665ddaf00bec29daf78cfadcfe7e4df/Packages/Status/Sources/Status/Editor/StatusEditorCoreView.swift#L97-L105

singingwolfboy:
> the maximum number of follow-up statuses is 5, be cautious when increasing it

Out of curiosity, why is there a limit at all? And what should we be cautious about when increasing the limit?

thai-d-v:
@singingwolfboy To prevent this feature from becoming a spam machine.

PR #1751, opened 2023-12-17
Update EmojiText to 3.2.0

### LINKED ISSUE #1738: Bug: Some custom emoji cause app to crash
## Environment:

 - OS: iOS 17.2
 - IceCubesApp version: 1.19.4 (latest as of 2023-12-13)

## Description

The app crashes when a post that makes heavy use of custom emoji is about to be displayed (shortly before it becomes visible in the timeline, or when accessing it directly via "Open in IceCubes" feature in Safari).

The crash currently prevents me from opening IceCubes, since it crashes immediately as soon as I open the app (since it wants to display the post...).

I don't know what exactly causes the problem or how it can be reproduced precisely, but here's a post which crashes the app: https://mastodon.de/@Test123/111573015901745646 (perhaps it's also another post from that thread)

## Related Issues

 - [x] Search that this bugs don't already exist before creating it.

divadretlaw:
Fixes #1738

Tested with iPhone 11, iOS 17.2 (oldest iPhone I have available at the moment) and with the mentioned posts in #1738 it doesn't crash anymore.

### DIFF
--- a/IceCubesApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/IceCubesApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -14,8 +14,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/divadretlaw/EmojiText",
       "state" : {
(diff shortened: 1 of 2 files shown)


Dimillian:
Thanks!

Issue #1746, opened 2023-12-16
Feature request: Improve rendering of App Store links


sh95014:
Example https://mstdn.social/@shaundon/111569283248192215 currently renders like a website preview, always slicing off parts of the app icon and using far more vertical space than necessary:

![IMG_0531](https://github.com/Dimillian/IceCubesApp/assets/95387068/0e8fb32b-23ef-4abb-8017-65d1ee7c2bdb)

Suggest specializing it to something like:

![IMG_0533](https://github.com/Dimillian/IceCubesApp/assets/95387068/3bae1e72-79c2-49eb-b129-1e80bfed5300)

Let me know if you're interested and I can look into a fix.

Dimillian:
I don't know. I think I like all links card to be displayed the same way but I'm open to enhancements.

PR #1756, opened 2023-12-17
AppStore Link Preview

sh95014:
A new custom layout for links that are known to be associated with square artwork (icons), such as App Store, Apple Music, and Spotify for now. This avoids the clipping caused by the default layout, and is more economical with vertical space without sacrificing content. Fixes https://github.com/Dimillian/IceCubesApp/issues/1746

Sample:
截圖 2023-12-17 下午12 17 37


Packages/Status/Sources/Status/Row/Subviews/StatusRowCardView.swift
```diff
@@ -114,4 +82,89 @@ public struct StatusRowCardView: View {
     }
     .buttonStyle(.plain)
   }
+
+  @MainActor
```
Dimillian, line 86: FYI you don't need `MainActor` but you should use `ViewBuilder` so you can remove the `Group`

Dimillian:
I like it! Thanks

Packages/Status/Sources/Status/Row/Subviews/StatusRowCardView.swift
shuang886: I do need to learn how to use that…  :)

PR #1764, opened 2023-12-19
fix: `AddAccountView`

thai-d-v:
The first impression is crucial, so we should enhance the `AddAccountView` as it's the first view a new user will see.

image


Dimillian:
Much better, thanks!

PR #1809, opened 2024-01-01
Fix settings done button logic

NSHaoSong:
With the new customizable tab bar, "Settings" can be rendered directly as the root view of a tab bar entry instead of as a "modal" presented from the root view. The "Done" button to dismiss the modal should be hidden if `isModal` is `false`.

### Steps to reproduce the bug

- Open "Settings > Tab Customizations", change one of the tab to "Settings"
- Switch to "Settings" tab

### Before fix

before

### After fix

after

### DIFF
--- a/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
+++ b/IceCubesApp/App/Tabs/Settings/SettingsTab.swift
@@ -53,7 +53,7 @@ struct SettingsTabs: View {
       .navigationBarTitleDisplayMode(.inline)
       .toolbarBackground(theme.primaryBackgroundColor.opacity(0.50), for: .navigationBar)
       .toolbar {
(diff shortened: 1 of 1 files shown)


NSHaoSong:
@Dimillian I am testing my changes on iPad, and I found the secondary column button is shown when settings are presented in "modal" (from 'Profile' > '...' button > 'Settings'). 

Is this expected?

Screenshot 2024-01-01 at 8 57 46 AM

Dimillian:
Ohhh no it's not. Should not be shown when presented as a modal. You could add it to your PR.

NSHaoSong:
@Dimillian Added a 2nd commit to hide the secondary column button in modal:  https://github.com/Dimillian/IceCubesApp/pull/1809/commits/898f4b634c0585bd366349da60b276e0954dd7fd

PR #1817, opened 2024-01-03
Update Ukrainian localization

pshadov:
## Descripotion
As a native Ukrainian speaker, I noticed that some texts are inconsistent or not translated.

## Changelog
- ⚠️ removed unused string `account.detail.about`. Please let me know if it is used somewhere, and I missed it somehow. I noticed that for editing your profile, `account.edit.about` is used, and when you open the profile page (yours or other users) - there is no "About" label about their info;
- ⚠️ removed stale `iPhone` string;
- removed the `parent view for EditTagGroupView` text from Localizable;
- translated newer texts;
- reviewed some older texts and fixed a couple of inconsistencies (e.g., in some places "Post" is translated as "Допис" (Dopys) and in others as "Публікація" (Publicatsia)).


Dimillian:
THanks!

PR #1819, opened 2024-01-03
shift the square.and.pencil icon so that the square is visually centered in the background rectangle

sh95014:
before:
before

after:
after

Please accept, this is driving me a little nuts.  😊


Dimillian:
Same lol

PR #1823, opened 2024-01-05
feat: Supports quickly switching accounts with a swipe-up gesture on …

dploeger:
…the tab bar on iOS

Like I said on Mastodon, this is my first dabble in SwiftUI and Swift, so here be dragons (probably).

It implements switching the current account by simply swiping up on the tab bar. I'm open to suggestions.

### DIFF
--- a/IceCubesApp/App/Main/AppView.swift
+++ b/IceCubesApp/App/Main/AppView.swift
@@ -10,6 +10,7 @@ import RevenueCat
 import Status
 import SwiftUI
 import Timeline
+import Models
 
 @MainActor
 struct AppView: View {
@@ -77,6 +78,27 @@ struct AppView: View {
           .toolbarBackground(theme.primaryBackgroundColor.opacity(0.50), for: .tabBar)
       }
     }
+    .gesture(DragGesture().onEnded({ value in
+        if value.translation.height < 0 {
+            var found = false
+            var accountToSet : AppAccount = appAccountsManager.availableAccounts[0]
+            for account in appAccountsManager.availableAccounts {
+                if found {
+                    accountToSet = account
+                    break
+                }
+                if account.id == appAccountsManager.currentAccount.id {
+                    found = true
+                }
+            }
+            var transation = Transaction()
+            transation.disablesAnimations = true
+            withTransaction(transation) {
+              appAccountsManager.currentAccount = accountToSet
+              HapticManager.shared.fireHaptic(.notification(.success))
+            }
+        }
+    }))
     .id(appAccountsManager.currentClient.id)
     .withSheetDestinations(sheetDestinations: $appRouterPath.presentedSheet)
   }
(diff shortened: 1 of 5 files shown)


IceCubesApp/App/IceCubesApp-release.entitlements
```diff
@@ -2,22 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>aps-environment</key>
```
Dimillian, line 5: This should probably not be commited

IceCubesApp/App/IceCubesApp.entitlements
```diff
@@ -2,22 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>aps-environment</key>
```
Dimillian, line 5: Same

IceCubesApp/App/IceCubesApp-release.entitlements
dploeger: Arrgggh. And here I am trusting XCode with Git commands. Sorry.

Dimillian:
It's working! Now I wonder why not a long tab to bring the account selector too?

dploeger:
I've removed the long tap. Should I re-add it?

Dimillian:
I'm wondering if the swipe should swap the account or present the account selector sheet. I think maybe present the account selector sheet?

dploeger:
Maybe we could do both. The swipe is (or "feels") like a quick gesture, used for doing stuff swiftly, while the long press is for something more slower.

So maybe: long press for account selector sheet and swipe for switching the account?

Dimillian:
Sound good to test

dploeger:
Could you please point me to the new code where you open the compose sheet in a tab? I'm not really sure how to trigger opening a sheet in SwiftUi.

Dimillian:
Hmm it won't be easy as the app account selector is not globally available. I would need to refactor a few stuff. For now it's in `AppAccountsSelectorView` and there is no way to trigger it easily. It's bundled with the rounded avatar button basically.

dploeger:
Ah account. Let me see, if I can manage the refactoring...

dploeger:
Done. I've also rebased and saw how you implemented the sheet of the New button. Maybe the accountsSelectorView can be implemented the same way, but please just take a look.

Dimillian:
I'm testing it a bit, and I find that a tad confusing. Not sure this is the right direction :/

dploeger:
Yeah, I'm with you on that. The code needs to look more beautiful.

Dimillian:
> Yeah, I'm with you on that. The code needs to look more beautiful.

I was thinking about the UX, too. It's hard to discover. The TabView from SwiftUI is not really super extensible.

dploeger:
That's true, but I think the flick gesture to switch accounts is a nice UX, thought it has to be documented because, yes, it's not obvious. But in day-to-day usage it's pretty nice. I could live without the long-press gesture tbh.

PR #1827, opened 2024-01-05
Turkish content is improved & added missing ones.

egesucu:
* Some turkish words were directly copy-pasted
* Some needed rework.


egesucu:
Quick note: Text("") produces empty string on String catalog.
To dismiss this, we can change them to Text(verbatim: "") as specified in [SwiftUI Doc](https://developer.apple.com/documentation/swiftui/text)

PR #1834, opened 2024-01-07
Feature: Use tagGroup icon in timeline quick access feature

Lilfaen:
I love the new timeline quick access feature and use it all the time.
Until now, tagGroups used a generic tag symbol, even though they have their own symbol. 
Before: 

After:

What do you think?

### DIFF
--- a/IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
+++ b/IceCubesApp/App/Tabs/Timeline/TimelineTab.swift
@@ -94,7 +94,7 @@ struct TimelineTab: View {
         lastTimelineFilter = newValue
       }
       switch newValue {
-      case let .tagGroup(title, _):
+      case let .tagGroup(title, _, _):
         if let group = tagGroups.first(where: { $0.title == title}) {
           selectedTagGroup = group
         }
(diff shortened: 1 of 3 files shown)


Dimillian:
So by doing that, you're wiping the existing user data as a change to a nonoptional in Codable, which will make the AppStorage fail to decode it. 
What could work is to make the new enum value an optional one. And to encode/decode it as optional. In the UI, just fallback to the existing default tag icon for the previously saved tag group. 

I would love this change if we could keep existing user data!

Lilfaen:
> So by doing that, you're wiping the existing user data as a change to a nonoptional in Codable, which will make the AppStorage fail to decode it. What could work is to make the new enum value an optional one. And to encode/decode it as optional. In the UI, just fallback to the existing default tag icon for the previously saved tag group.

You're right, thanks! I fixed it, it should now successfully decode the previously saved pinned timelines. 
I'm not sure though how to update the icon without deleting the pin and adding it again.

Dimillian:
@Lilfaen Yes there is probably a way if we match the title but let's do do that. Testing and merging. Thanks!

PR #1844, opened 2024-01-08
Replace profile navigation bar back button

Matt54:
How do you feel about a consistent back button on the profile?

Before:
RocketSim_Screenshot_iPhone_15_6 1_2024-01-08_07 06 49

After:
RocketSim_Screenshot_iPhone_15_6 1_2024-01-08_07 11 01


Dimillian:
I agree it's better, but I'm not for adding custom code to handle built-in iOS features. What about visionOS or macOS? 
Would be better if SwiftUI had something built in to do that.

Matt54:
That makes sense. Hopefully Apple will give us something SwiftUI-friendly to clean that up over time..

Beautiful codebase btw!

Dimillian:
Thanks!

PR #1877, opened 2024-01-13
Added toggleable button to show/hide poll results

eric-Is-Me:
- Placed the "Send Vote" and "Show/Hide Results" buttons in an HStack
- Added english localizable strings for show/hide results
- Changed "Send Vote" button style to Prominent Bordered in order to contrast with the Bordered style of the new "Show/Hide Results" button
- Changed the showResults property in the view model from a computed var to a stored var since this implementation wants to toggle as well as observe changes to this property. It is extending more responsibility to this property but I feel like it makes sense to use it this way?

attached demo pics~
Screenshot 2024-01-12 at 7 36 13 PM
Screenshot 2024-01-12 at 7 36 01 PM

### DIFF
--- a/IceCubesApp/Resources/Localization/Localizable.xcstrings
+++ b/IceCubesApp/Resources/Localization/Localizable.xcstrings
@@ -67603,6 +67603,17 @@
         }
       }
     },
+    "status.poll.hide-results" : {
+      "extractionState" : "manual",
+      "localizations" : {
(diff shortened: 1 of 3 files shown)


Dimillian:
Loving it! I wanted to rework polls but this will actually do it! I'll make a few adjustments and merge it. Thanks!

Issue #1878, opened 2024-01-13
Feature request: TR Localization Update


egesucu:
Update TR Language

PR #1880, opened 2024-01-13
TR Localization Update #1878

egesucu:
#1878


egesucu:
The removal part was marked by Xcode as "stale", meaning it was not being used anymore.

PR #1883, opened 2024-01-13
Correct platform name in display settings depending on device

infinitepower18:
Currently, the "enable secondary column" setting on the Mac is under an "iPad" section in display settings:

image

This PR fixes that by returning the correct name based on the device (based on Xcode 15's ability to vary strings based on device).


Dimillian:
Nice!

PR #1888, opened 2024-01-14
Using AsyncButton from button kit to improve following and related buttons

Dean151:
- Following will now show an overlay spinner when loading
- If an error occurs, it'll shake

- Boost and notification toggle will disable when loading
- If an error occurs, they'll shake


Dimillian:
Amazing! thanks!

PR #1895, opened 2024-01-17
Resolve #359 Optional Missing Alt-Text warning

### LINKED ISSUE #359: Feature request: optional Missing Alt-Text warning
Some instances require media to have alt-text, so a warning before posting a toot that has an attachment but no alt text would be useful. Could be made optional in the settings.

Lr-K:
Resolves https://github.com/Dimillian/IceCubesApp/issues/359

- Add toggle in settings to require alt text (default off)
- If setting is enabled, posting show an error if any attached media is missing alt text

Design decisions:
- Chose to surface an error instead of disabling posting, to make it obvious why the post is failing. A tooltip goes against current design decisions and isn't clear on all platforms.
- The setting is default off to minimise impact to existing users.

Setting in content settings

Toggle setting in content settings

Error displayed to user if attempting to post without all the necessary alt text

Error displayed to user if attempting to post without all the necessary alt text (if setting is toggled on)

Tested on:
- iPadOS Simulator with photos and videos

### DIFF
--- a/IceCubesApp/Resources/Localization/Localizable.xcstrings
+++ b/IceCubesApp/Resources/Localization/Localizable.xcstrings
@@ -41563,7 +41563,115 @@
     },
     "settings.content.require-alt-text" : {
       "localizations" : {
+        "be" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Require Alt Text to Post"
+          }
+        },
+        "ca" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Require Alt Text to Post"
+          }
+        },
+        "de" : {
+          "stringUnit" : {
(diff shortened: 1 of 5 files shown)


IceCubesApp/Resources/Localization/Localizable.xcstrings
```diff
@@ -41668,6 +41668,16 @@
         }
       }
     },
+    "settings.content.require-alt-text" : {
```
Dimillian, line 41671: Unfortunately, due to how String catalog works, the default Enligh has to be added to all languages.

Lr-K: Thanks, I thought it used a fallback. I have added the English translation to all the other languages now.

PR #1901, opened 2024-01-18
Fix `StatusRowCardView` layout.

thai-d-v:
I've fixed the layout to make it more tidy, balanced and consistent:
- make the description font equal to the URL font
- use conventional numbers of avatar's configs for padding and corner radius
- set the title's line limit to 1

image


Dimillian:
Let me sit on it a bit, in one case you could read the title fully hmm

thai-d-v:
Yes, a tidy layout and concise content are what I'm thinking about the preview. I think a poll on Mastodon about this is maybe a good idea.

PR #1904, opened 2024-01-19
Fix Preview Image Size of `StatusRowCardView`.

thai-d-v:
Nowadays, many websites have very well-designed preview images. We should take care of them because they convey information and make the app UI look very nice. 😁

image

### DIFF
--- a/Packages/Models/Sources/Models/Card.swift
+++ b/Packages/Models/Sources/Models/Card.swift
@@ -10,6 +10,8 @@ public struct Card: Codable, Identifiable, Equatable, Hashable {
   public let description: String?
   public let type: String
   public let image: URL?
+  public let width: CGFloat
+  public let height: CGFloat
 }
 
 extension Card: Sendable {}
--- a/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
+++ b/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
@@ -158,6 +158,8 @@ public struct StatusRowCardView: View {
 }
 
 struct DefaultPreviewImage: View {
+  @Environment(Theme.self) private var theme
+  
   let url: URL
   let originalWidth: CGFloat
   let originalHeight: CGFloat
@@ -165,7 +167,8 @@ struct DefaultPreviewImage: View {
   var body: some View {
     _Layout(originalWidth: originalWidth, originalHeight: originalHeight) {
       LazyResizableImage(url: url) { state, _ in
-        Rectangle().fill(Color.gray)
+        Rectangle()
+          .fill(theme.secondaryBackgroundColor)
           .overlay {
             if let image = state.image {
               image.resizable().scaledToFill()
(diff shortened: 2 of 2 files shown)


Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
```diff
@@ -91,28 +84,48 @@ public struct StatusRowCardView: View {
...
+
+    var body: some View {
+      LazyResizableImage(url: url) { state, proxy in
+        let proxyWidth = imageWidthFor(proxy: proxy)
+        if let image = state.image,
+           let cgImage = state.imageContainer?.image.cgImage
+        {
+          image.resizable().scaledToFit()
+            .onAppear {
+              self.width = proxyWidth
```
Dimillian, line 104: The thing I don't like about that is that it's costly. It'll create stutters as as the image appears. I wish we could solve that differently, without relying to onAppear.

Dimillian: I don't think I can let that in. I agree it's better and how it's supposed to look but with SwiftUI List it creates performance issues. As stuff get resized while scrolling.

Dimillian: Actually the API state what the should have `width` and `height` in `Card`. Maybe you could add this and we could calculate its size statically. We could even show the square preview if it's a square.

thai-d-v:
I did it.

1. calculate the preview image size statically
2. fix the text spacing when there is no preview image

Some links have ratio/size mismatches between API data and actual images, for example, [Changelog Podcast](https://changelog.com/podcast/566), but look still better than before.

Packages/Models/Sources/Models/Card.swift
```diff
@@ -10,6 +10,8 @@ public struct Card: Codable, Identifiable, Equatable, Hashable {
   public let description: String?
   public let type: String
   public let image: URL?
+  public let width: CGFloat
+  public let height: CGFloat
```
Dimillian, line 14: Are you confident it's not optional? Maybe it should so we don't crash the whole feed.

Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
```diff
@@ -173,3 +156,54 @@ public struct StatusRowCardView: View {
...
+              image.resizable().scaledToFill()
+            }
+          }
+      }
+      .accessibilityHidden(true) // This image is decorative
+      .clipped()
+    }
+  }
+
+  private struct _Layout: Layout {
```
Dimillian, line 180: Very interesting! I've never used the Layout API! I need to read about it.

Packages/Models/Sources/Models/Card.swift
thai-d-v: yes, I've read the API doc

image

Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
thai-d-v: We can further stabilize the timeline layout with it. 😁

Dimillian: Yes, I need to read about it! I feel like this could help for more stuff!

PR #1924, opened 2024-01-24
Add shimmering effect to loading placeholders

marcelvoss:
I was a bit bothered by the lack of animation during the loading phase of timelines between screens. While I think the `.redacted(.placeholder)` option is a good choice, I think a tiny bit of shimmering makes it a bit less static and shows that something is still loading.

Let me know what you think,
Cheers


Dimillian:
I'm sorry you spent time on this. I recently removed them because it cause a small performance overhead that is visible in some transition unfortunately.

marcelvoss:
Ah, all good. Well 🤣

PR #1928, opened 2024-01-28
Fix timeline media size

thai-d-v:
- use `Layout` to eliminate `SceneDelegate` and other dependencies
- roll back to grid layout for posts with multiple media items because it is more intuitive
- use a safe number `450` for maximum height because can not detect the height of the scroll view, but it works very well for the layout

image
image
image
image

### DIFF
--- a/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
+++ b/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
@@ -132,11 +132,11 @@ private struct MediaPreview: View {
               .frame(width: displayData.isLandscape ? imageMaxHeight * 1.2 : imageMaxHeight / 1.5,
                      height: imageMaxHeight)
               .overlay(
-                RoundedRectangle(cornerRadius: 4)
+                RoundedRectangle(cornerRadius: 10)
                   .stroke(.gray.opacity(0.35), lineWidth: 1)
               )
           } else if state.isLoading {
-            RoundedRectangle(cornerRadius: 4)
+            RoundedRectangle(cornerRadius: 10)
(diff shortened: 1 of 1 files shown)


Dimillian:
This is awesome but I really want to keep the carousel for the previews. This is better IMO and also future-proof as some instances support more than 4 media (like Threads soon)

thai-d-v:
Of course, the carousel is back, I just changed the corner radius to make it consistent.

Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
```diff
@@ -35,33 +31,6 @@ public struct StatusRowMediaPreviewView: View {
   private var scrollBottomPadding: CGFloat? = 0
 #endif
 
-  var availableWidth: CGFloat {
```
Dimillian, line 38: This is amazing that this is gone!

Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
```diff
@@ -517,3 +386,138 @@ struct WrapperForPreview: View {
...
+  var body: some View {
+    if let url = attachment.url {
+      _Layout(originalWidth: originalWidth, originalHeight: originalHeight, maxSize: maxSize) {
+
+        Group {
+          RoundedRectangle(cornerRadius: 10).fill(Color.gray)
+            .overlay {
+              switch attachment.supportedType {
+              case .image:
+                LazyImage(url: attachment.url) { state in
```
Dimillian, line 418: Swapping to `LazyResizableImage` so we get the processors

PR #1929, opened 2024-01-29
Improve `StatusPollView`

thai-d-v:
1. use `Layout` to remove `GeometryReader`
2. fix inconsistent bars' height
3. move the button image out of the bar to make it clear which bar is selected
4. use relative percent for bars' width
    - more space-efficient, show the relative comparison between bars better
    - the "absolute percent" is already shown by the text on the right
5. refactor code and tweak layout to make it more consistent with other views

image


thai-d-v:
I hope this is OK.

image

Dimillian:
Nice!

PR #1964, opened 2024-02-12
update zh_Hant

sh95014:


sh95014:
not needed since you took https://github.com/Dimillian/IceCubesApp/pull/1968

Issue #1932, opened 2024-01-29
Bug: "Show Sensitive Content" button text can be illegible if background is dark


sh95014:
## Environment:

 - OS: 17.3
 - IceCubesApp version: 1.10.28

## Description

1. Settings > Display Settings > Tint Color = one of the darker colors, such as the darkest purple
2. View a post with a sensitive image, such as https://mastodon.xyz/@xkcd/111825444296281157

![IMG_6521973C64AE-1](https://github.com/Dimillian/IceCubesApp/assets/95387068/4da1623a-d7d1-4afc-8964-880bea5a9be3)

## Related Issues

 - [X] Search that this bugs don't already exist before creating it.

PR #1965, opened 2024-02-12
select a contrasting color for label of "show sensitive content" button

sh95014:
fix https://github.com/Dimillian/IceCubesApp/issues/1932

截圖 2024-02-11 下午5 34 57
截圖 2024-02-11 下午5 34 46


Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
```diff
@@ -245,6 +246,29 @@ struct BlurOverLay: View {
     default: false
     }
   }
+  
+  // return either labelColor or primaryBackgroundColor, whichever contrasts better against
+  // the specified background color
+  private func contrastingColor(against backgroundColor: Color) -> Color {
```
Dimillian, line 252: This should probably live in Theme and be cached, this is a bit performance consuming to have this in the view

Dimillian:
This is really nice thanks!

PR #1967, opened 2024-02-12
Limit height of images in previews to screen height

sh95014:
limit available height to 60% of screen/window height so that the entire card would approximately fit

basically https://github.com/Dimillian/IceCubesApp/pull/1675, but for link previews instead of plain images.

### DIFF
--- a/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
+++ b/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
@@ -225,13 +225,14 @@ public struct StatusRowCardView: View {
 
 struct DefaultPreviewImage: View {
   @Environment(Theme.self) private var theme
+  @Environment(SceneDelegate.self) private var sceneDelegate
   
   let url: URL
   let originalWidth: CGFloat
   let originalHeight: CGFloat
 
   var body: some View {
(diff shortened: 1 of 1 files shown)


Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowCardView.swift
```diff
@@ -225,13 +225,14 @@ public struct StatusRowCardView: View {
...
   @Environment(Theme.self) private var theme
+  @Environment(SceneDelegate.self) private var sceneDelegate
   
   let url: URL
   let originalWidth: CGFloat
   let originalHeight: CGFloat
 
   var body: some View {
-    _Layout(originalWidth: originalWidth, originalHeight: originalHeight) {
+    _Layout(originalWidth: originalWidth, originalHeight: min(originalHeight, sceneDelegate.windowHeight * 0.6)) {
```
Dimillian, line 235: Need to think about this one, I don't think we should rely on scenelDelegate but we should provide an absolute height instead. Do you have a link to a problematic preview?

sh95014: https://mastodon.social/@hikingdude/111843640004810123 (on an iPad)

PR #1985, opened 2024-02-19
Handle edge cases for `StatusRowCardView`

thai-d-v:
### Problem

I am trying to solve the problem at #1967. I don't limit the preview image height in the first place (#1904) because I don't know how to handle its display. Two critical aspects of `StatusRowCardView` are __card layout__ and __preview content.__

- Fill Image: keep the card layout but cut off the content
- Fit Image: keep the content but break the card layout with two dangling spaces (left/right)

### Solution

1. Using a Fit Image to keep the content of __intentionally well-designed__ preview images
2. Use a blurred Fill Image in the background to:
    - keep the card layout
    - give spaces nuance colors to keep the card display __coherent__
3. max height is `450` (same as media max height)

_(high resolution image, you should zoom in to see the details)_
image


Dimillian:
I think I like this solution, I'll do some testing around and see. Thanks!

Issue #1845, opened 2024-01-08
Bug: Swipe actions are overlapped with images 


pshadov:
## Environment:
 - OS:  iOS 17
 - IceCubesApp version: 1.10.16

## Description
1. The user has multiple images on the Post (Status)
2. Try to swipe left to see the actions

Result:
The images are not clipped and overlap swipe actions.
![Simulator Screenshot - iPhone 13 mini - 2024-01-08 at 12 32 59](https://github.com/Dimillian/IceCubesApp/assets/6460917/a3b0c32f-c3ca-4762-a40c-d5a0dbe32472)

## Related Issues
I searched that this bug didn't already exist before creating it.

alexsereno:
https://github.com/Dimillian/IceCubesApp/assets/2569805/c700f969-ab14-43d0-8fae-33e689ef32fc

mdhsieh:
I'm guessing this can be fixed with a z-index, but I'll need to look more into this.

PR #2000, opened 2024-03-08
Swipe actions overlapped with images #1845

mdhsieh:
### Environment:
- OS: iOS 17.3.1
- IceCubesApp version: 1.10.33
- Issue #1845

### Issue:
https://github.com/Dimillian/IceCubesApp/issues/1845 Multiple images overlap left swipe actions.

### Causes:
In `StatusRowMediaViewPreview`, the [`scrollClipDisabled()`](https://www.hackingwithswift.com/quick-start/swiftui/how-to-disable-scrollview-clipping-so-contents-overflow) modifier allows multiple images to overflow outside the scrolling area. However from my understanding it was designed to flow on top of other content.

### Solution:
Replace `scrollClipDisabled()` with negative horizontal padding, and the same amount of content margin. That way we have the same visual effect as before, but there's no overlap problem.

Tested on iPhone and iPad 17.0

Demo video:

https://github.com/Dimillian/IceCubesApp/assets/52055110/97f5bbe6-9281-4af7-9952-e03f7cfcdab0

### DIFF
--- a/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
+++ b/Packages/StatusKit/Sources/StatusKit/Row/Subviews/StatusRowMediaPreviewView.swift
@@ -67,7 +67,11 @@ public struct StatusRowMediaPreviewView: View {
           }
           .padding(.bottom, scrollBottomPadding)
         }
-        .scrollClipDisabled()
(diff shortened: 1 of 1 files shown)


Dimillian:
I'm closing this for now, I agree that it fixes it, but it's cutting the effect instead of visually showing infinite scrolling.  I would prefer to find something better, if possible.

PR #2001, opened 2024-03-09
Added a "Moved To" Button to accounts that moved to other instances

343max:
It's close to impossible to see when a user moved to a different instance.

I added a button that allows the user to navigate to the new profile.

If this can be merged then I'd love to change the "Follow" button to "New Profile" so the user doesn't try to follow a closed account (which doesn't work anyways)

image


Dimillian:
That's actually super nice!

PR #2006, opened 2024-03-16
Feature: RSS Reader

thai-d-v:
### The Case for RSS in IceCubesApp, and Other Mastodon Clients

RSS has been a very popular way to distribute content for a very long time. RSS feeds content that is high-quality, focused, and very specialized. If you want to follow people but only on professional topics, following them on social networks is a good choice, but not ideal since their timeline may contain a lot of other topics. If their websites/blogs have RSS, following RSS feeds can be a better move.

RSS content usually serves as an overview version of content on websites/blogs. That's why RSS items can be well fit on Mastodon's timelines. NetNewsWire's Brent Simmons also talks about this. [On Mastodon Support in NetNewsWire.](https://inessential.com/2023/12/17/on_mastodon_support_in_netnewswire.html)

Due to Mastodon's developing status, the network isn't yet big enough for users accustomed to abundant content on platforms like Twitter and Facebook. Consequently, there is a strong demand for expanding the network. 

RSS is highly distributed and doesn't require users to create or manage an account, making it even more distributed than Mastodon. As a result, connecting to RSS is less challenging than connecting to Thread or Twitter. 

Overall, connecting Mastodon to RSS provides great benefits for users of both Mastodon and RSS.

### How do I make RSS fit in IceCubesApp?

RSS readers usually design item lists in a very classic way, with a title and a small avatar. It is pretty far from the UI taste of social network users. So the way I make it more user-friendly is:

1. Turn the item list into an RSS timeline.
2. RSS Item appears to look like a Mastodon post:
    - always has avatar, title, date, and summary
    - sometimes has a preview image
3. You can interact with RSS items in a social network style:
    - post the RSS item on Mastodon
    - share with other apps (and Mastodon)
    - mark it as read/unread
4. Users will read full content, with a complete browsing experience right inside IceCubesApp.
5. Onboarding users with natural instructions

### Technical Summary

1. Design patterns are **VM** and **MVVM**, depend on the complexity of UI interaction
2. **RSS tab** connects to the app like other tabs and can be **customized**
3. RSS is organized as **a local package** with a minimal interface
4. use parser **RSParser** of NetNewsWire
5. 100% compatible with **Swift 6 Concurrency**
6. use **Core Data** because it works very nicely, SwiftData doesn't
7. use **Regex** because it is fast, SwiftSoup isn't

#### 1. Onboarding

https://github.com/Dimillian/IceCubesApp/assets/46838577/9b4acde8-2288-4967-9deb-e07110d36b1e

#### 2. RSS Managing

- show/ hide feeds
- add new feeds

https://github.com/Dimillian/IceCubesApp/assets/46838577/f0d0901b-113f-4d6d-b6ca-e30043c9c959

#### 3. Interactions

- read the full content
- mark items read/unread
- post on Mastodon

https://github.com/Dimillian/IceCubesApp/assets/46838577/b9d19fae-d6fb-4129-a54c-a7ff3ed1aee1


thai-d-v:
@Dimillian In case you missed it, what are your thoughts on this?

Dimillian:
First thanks a lot for this. I'm actually amazed you added this feature. Now as for Ice Cubes as a product, I'm not sure a RSS reader is a good fit. Might be a good idea to spin another app
There is already so many features related to Mastodon I don't see maintaining a whole other product right within ice cubes unfortunately.

thai-d-v:
Personally, of course, I see RSS as a good fit. But objectively, for your user base, I'm not sure, too. That's why I encourage you to give it a try. 

If it's NOT a good fit, users can replace it with another tab, or you can completely remove it just by removing the package and deleting a few lines of code in the main app.

But if users LOVE it enough, IceCubesApp will have an advantage in attracting many more users. At that point, I think RSS is worth the maintenance.

At the end of the day, as the owner of the project, it's up to you to decide what's best for your user base.

PR #2011, opened 2024-03-22
Add SwiftPolyglot to validate localizations

roddymunro:
I thought it would be a good idea to add my new script, [SwiftPolyglot](https://github.com/appdecostudio/SwiftPolyglot), as a PR check for IceCubes. The script simply parses all `.xcstrings` in the repository and ensures that there are translations present for every key in every localization that the app supports.

I'm adding this without the `--errorOnMissing` flag, as I noticed there's a few translations missing. This means the step will "pass", but you will be able to see all the warnings by viewing the GitHub action's output. Annotations will also be added to the `.xcstrings` files missing translations.

Here's the output of the script: https://github.com/roddymunro/IceCubesApp/actions/runs/8383388923/job/22959006951


Dimillian:
That's super nice, let me run and see what is missing so we can merge and enable it

PR #2056, opened 2024-05-05
Update EU localisation

xabirequejo:
Newly added strings


Dimillian:
I'm sorry but I might have updated the string a bit already :/

PR #2065, opened 2024-05-10
Implement Apple Translate

Havhingstor:
Instead of just their instance and DeepL, the user has now the option between their instance, DeepL and Apple Translate. The user's previous choice (DeepL on / off) is reflected in the first value of the new setting and the old always_use_deepl App-Storage Key is deleted.
The option is only available on iOS 17.4 or later and not on Mac or Vision Pro. If the user is on an unsupported platform but the option to use Apple Translate is set, it's reset to use the instance.

This also includes a couple of fixes that aren't necessary for Apple Translate, but related:
- The background of the "Auto detect language" button is now consistent with all the other rows
- All loaded translations are removed if the translation type is switched
- A warning is shown to the user if they've switched away from using DeepL, but there's still an API key stored
- The spelling of "DeepL API Key" is unified, previously the "Key" was sometimes lowercase.

Tested on iPhone, iPad, Mac & visionOS Simulator

### DIFF
--- a/IceCubesApp/Resources/Localization/Localizable.xcstrings
+++ b/IceCubesApp/Resources/Localization/Localizable.xcstrings
@@ -26160,6 +26160,23 @@
         }
       }
     },
+    "enum.translation-type.use-server-if-possible" : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "de" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "Instanz"
+          }
+        },
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Instance"
+          }
+        }
+      }
+    },
     "env.poll-vote-frequency.multiple" : {
       "extractionState" : "manual",
       "localizations" : {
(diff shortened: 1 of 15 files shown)


IceCubesApp/Resources/Localization/Localizable.xcstrings
```diff
@@ -26160,6 +26160,45 @@
         }
       }
     },
+    "enum.translation-type.apple" : {
```
Dimillian, line 26163: FYI, when adding a new translation, I'm OK with using English localization as the key. It makes the process much simpler as we don't have to add the default english localization to every languages.

Dimillian:
I wonder if it's sime to remove the pro key I provide in DeepLClient. It's a nice faillback when instance translate is not available, but it's expensive.

IceCubesApp/Resources/Localization/Localizable.xcstrings
Havhingstor: I can revert the one commit that changes that for Apple Translate and DeepL

Havhingstor:
And I can also remove the default DeepL client.

Dimillian:
> And I can also remove the default DeepL client.

Still thinking about it, I need to review the case where the app use this instead of the instance service. I think it's when the instance service is missing and when the user doesn't have a free key. But there is a lot of usage.

Packages/StatusKit/Sources/StatusKit/Row/StatusRowView.swift
```diff
@@ -219,6 +233,23 @@ public struct StatusRowView: View {
       StatusDataControllerProvider.shared.dataController(for: viewModel.finalStatus,
                                                          client: viewModel.client)
     )
+    .alert("alert.translation-error.deepl", isPresented: $viewModel.deeplTranslationError) {
```
Dimillian, line 236: What I was saying about localization is that for the new stuff you want to put in, you can put the English inline directly, it'll become the key and also the default localization for every languages not providing a translation.

Packages/Env/Sources/Env/TranslationType.swift
```diff
@@ -0,0 +1,18 @@
...
+
+public enum TranslationType: String, CaseIterable {
+  case useServerIfPossible
+  case useDeepl
+  case useApple
+
+  public var description: LocalizedStringKey {
+    switch self {
+    case .useServerIfPossible:
+      "enum.translation-type.use-server-if-possible"
```
Dimillian, line 11: Same for this, if we don't have it in localization dic, use the English string directly.

Packages/Network/Sources/Network/DeepLClient.swift
```diff
@@ -12,20 +12,8 @@ public struct DeepLClient: Sendable {
     "https://api\(deeplUserAPIFree && (deeplUserAPIKey != nil) ? "-free" : "").deepl.com/v2/translate"
   }
 
-  private var APIKey: String {
```
Dimillian, line 15: I think it's fine to remove the app since it now defaults to the Apple translation service on iOS 17.4. We will see how the users react.

Dimillian:
Once it's done, all good I'll merge. Thanks!

Dimillian:
Just a thought: In StatusRowViewModel, if the instance or DeepL translation fails (because we now removed the app's own PRO key), what if we fallback to Apple translate (without even showing an error) for users on iOS 17.4, where the feature is available?

Havhingstor:
Sounds good! I'd still keep DeepL as the first fallback for the instance if the user has a key, though

Havhingstor:
I think I've transformed all the keys I've added to clear text strings, and I've also removed your API key, and made DeepL and the Instance fallbacks for each other and Apple Translate for both.

Dimillian:
Thanks!

Dimillian:
🚀

PR #2071, opened 2024-05-13
Resolve escaped characters in a status

Havhingstor:
Escaped characters are now returned to their original form for HTMLString.asRawText. That's important for translating posts.
See https://osna.social/@Mordoukna/112082755067296741


Packages/Models/Sources/Models/Alias/HTMLString.swift
```diff
@@ -61,7 +61,7 @@ public struct HTMLString: Codable, Equatable, Hashable, @unchecked Sendable {
           _ = text.removeLast()
           _ = text.removeLast()
         }
-        asRawText = text
+        asRawText = (try? Entities.unescape(text)) ?? ""
```
Dimillian, line 64: Should be `(try? Entities.unescape(text)) ?? text` right?

PR #2083, opened 2024-05-20
Disable hashtag navigation for avoiding redundant fetching

### LINKED ISSUE #2082: Bug: Redundant fetching for hashtag navigation
## Environment:

 - OS: 
iOS 14.4.1
 - IceCubesApp version: 
Version 1.10.40

## Description

Whenever the user navigates to a certain hashtag, if they're already on the view for that hashtag they still will be redirected to the same hashtag, navigating again which allocates more memory for the fetching and can lead to memory bloating. It is also redoing the fetch every time unnecessarily.

## Evidence

https://github.com/Dimillian/IceCubesApp/assets/77981728/00a12b0e-5aed-4d03-a792-41602a0fa383

I also ran this using simulator and instruments to see how it is affecting performance and memory.

image

## Related Issues

 - [x] Search that this bugs don't already exist before creating it.

juanyepesp:
I noticed that you can navigate to the same hashtag that you are currently at, which ultimately can lead to memory bloating and, if done too many times, can freeze or crash the app. To better understand the situation I am referring to, check the linked issue.

This hopefully fixes the issue, modifying the `Router` file inside the `Env` package. These modifications disable navigation from the same destination you are currently at, it can be either hashtags (case-insensitively), user profiles, etc. Hopefully I am not introducing any bugs or performance issues!

- fixes #2082

### DIFF
--- a/Packages/Env/Sources/Env/Router.swift
+++ b/Packages/Env/Sources/Env/Router.swift
@@ -27,7 +27,27 @@ public enum RouterDestination: Hashable {
   case notificationForAccount(accountId: String)
   case blockedAccounts
   case mutedAccounts
+
+  public static func == (lhs: RouterDestination, rhs: RouterDestination) -> Bool {
+    switch (lhs, rhs) {
+    case (.hashTag(let lhsTag, let lhsAccount), .hashTag(let rhsTag, let rhsAccount)):
+      return lhsTag.caseInsensitiveCompare(rhsTag) == .orderedSame && lhsAccount == rhsAccount
+    default:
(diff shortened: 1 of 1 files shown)


Packages/Env/Sources/Env/Router.swift
```diff
@@ -134,13 +135,25 @@ public enum SettingsStartingPoint {
 
   public var path: [RouterDestination] = []
   public var presentedSheet: SheetDestination?
+  public var currentDestination: RouterDestination?
```
Dimillian, line 138: You don't need that `path.last` will give you exactly what you need.

PR #2100, opened 2024-06-12
Full Xcode 16 supports + iOS 18 support

Dimillian:


koyakei:
Thank you. Im really waiting for it. @Dimillian

IceCubesApp/App/Report/ReportView.swift
```diff
@@ -44,6 +44,7 @@ public struct ReportView: View {
           ToolbarItem(placement: .navigationBarTrailing) {
             Button {
               isSendingReport = true
+              let client = client
```
TheRogue76, line 47: May i ask why this was needed?

Dimillian: I think it was one of the beta bug, I'll try to remove it.

Dimillian: Removed

PR #2107, opened 2024-06-16
Clip StatusRowView to prevent media from hiding swipe action buttons

### LINKED ISSUE #2106: Bug: Multiple embedded images hide swipe actions
## Environment:

 - OS: iOS 17.5.1
 
 - IceCubesApp version: 1.10.42
 

## Description

When a status contains multiple images, they overlap with the swipe action buttons, hiding them.

Simulator Screenshot - iPhone 15 - 2024-06-16 at 21 47 52

asendra:
## Description

When a status contains multiple images, they overlap with the swipe action buttons, hiding them.

Fixes #2106 

This PR clips the `StatusRowView` and switches the horizontal inset for horizontal padding.

At first, I only added the clipping. 

Unfortunately, due to the list inset, this made the content not go edge to edge, defeating what I guess was the desired effect of the feature. (See alternative solution for the result)

This is the reason why I also had to replace the list horizontal inset for the horizontal content padding.

## Before
before

 ## After
fix

 ## Alternative solution
alternative

### DIFF
--- a/Packages/StatusKit/Sources/StatusKit/Row/StatusRowView.swift
+++ b/Packages/StatusKit/Sources/StatusKit/Row/StatusRowView.swift
@@ -115,8 +115,9 @@ public struct StatusRowView: View {
           }
         }
       }
-      .padding(.init(top: isCompact ? 6 : 12, leading: 0, bottom: isFocused ? 12 : 6, trailing: 0))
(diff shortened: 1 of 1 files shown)


asendra:
Hey! Just curious, is this not considered a bug, or was the solution bad?

PR #2115, opened 2024-06-28
Remove space between Settings and ellipsis

infinitepower18:


infinitepower18:
@Dimillian why was this closed? Having the space makes it inconsistent with the rest of macOS which doesn't have the space between Settings and ellipsis.

Issue #2117, opened 2024-07-02
Improve french localization


pylapp:
Some wordings in use in the app for french language are not translated, and some can be improved.

PR #2118, opened 2024-07-02
Update french wordings (#2117)

pylapp:
Some wordings were not translated in french, others needed to be improved a bit.
This fix brings evolutions written by myself and tested locally on a real device (french guy with french as mother language).


pylapp:
@Dimillian Seems good for review 🍪

pylapp:
Merci surtout à toi pour cette superbe app !

PR #2130, opened 2024-07-16
Update Basque localisation

xabirequejo:
gradient


Dimillian:
I think we'll start using a third party to do translation because I can't merge or even see the diff here :/ It says everything changed.

PR #2139, opened 2024-07-22
Update EU localisation, take 2

xabirequejo:
Third party platform would be lovely.
I'm not sure what I'm doing because the file is so big that GitHub won't preview it, not even just the changes.
There is also this string in German, that I'm not sure I should be seeing: `%@ ist keine valide Instanz`


Dimillian:
I'll take a look.. The diff is weird indeed

PR #2142, opened 2024-07-23
Add context menu and draggability to attachment image views

fwcd:
This adds a context menu and the standard drag gesture to attachment images:

https://github.com/user-attachments/assets/c26d98b7-f6b6-4885-a879-3f493b139c1a


Dimillian:
Awesome! Thanks!

PR #2145, opened 2024-07-25
Replaced to async method for URLSessionWebSocketTask.receive().

kitwtnb:
Using MainActor properties in an unstructured task would cause build errors in Swift 6, so `URLSessionWebSocketTask.receive(completionHandler:)` was replaced with ` URLSessionWebSocketTask.receive() async throws`.

### DIFF
--- a/Packages/Env/Sources/Env/StreamWatcher.swift
+++ b/Packages/Env/Sources/Env/StreamWatcher.swift
@@ -36,16 +36,16 @@ import OSLog
     decoder.keyDecodingStrategy = .convertFromSnakeCase
   }
 
-  public func setClient(client: Client, instanceStreamingURL: URL?) {
+  public func setClient(client: Client, instanceStreamingURL: URL?) async {
     if self.client != nil {
       stopWatching()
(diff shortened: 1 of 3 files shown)


Dimillian:
This is cool thanks! Will try in a bit and merge!

Packages/Env/Sources/Env/StreamWatcher.swift
```diff
@@ -83,48 +83,42 @@ import OSLog
...
-              Task { @MainActor in
-                self.events.append(event)
-                self.latestEvent = event
-                if let event = event as? StreamEventNotification, event.notification.status?.visibility != .direct {
-                  self.unreadNotificationsCount += 1
-                }
-              }
+  private func receiveMessage() async {
+    do {
+      guard let message = try await task?.receive() else { return }
```
Dimillian, line 88: I don't think you should return here as it could break the next receiveMessage loop.

Dimillian:
This is weird, I've been testing it extensively and never received any message.

PR #2153, opened 2024-08-03
Collapse posts on tap (Apollo-style)

### LINKED ISSUE #1950: Feature request: Tap to Collapse Reply Threads
Search that this request don't already exist before creating it.

Would be fantastic to be able to tap a parent reply that would collapse all replies to that reply. Some posts promote a lot of discussion and become tedious to scroll through just to see the next post reply rather than a reply reply. Replies to replies for some controversial subjects could also be filled with arguing that's not relevant to the original post.

fwcd:
### Fixes #1950

With this PR threads are now collapsed instead of focused when tapped:

https://github.com/user-attachments/assets/6adfdf4f-3928-40f8-9daf-07a83006d31b

This is similar to how apps like Apollo, Mlem and other apps with hierarchical comments handle this. In its current state, the PR _replaces_ the focus-on-tap behavior for any indented replies, if desired, I could see a few ways of reconciling the old/current behavior with this though:

- Provide a focus action as a separate context menu
- Provide a setting for changing the "tap" action (i.e. focus or collapse)
  - Perhaps we could rename the "Swipe Actions" settings to a more general "Gestures" and put it there?

Would love to hear some thoughts!

### DIFF
--- a/Packages/StatusKit/Sources/StatusKit/Detail/StatusDetailView.swift
+++ b/Packages/StatusKit/Sources/StatusKit/Detail/StatusDetailView.swift
@@ -113,12 +113,15 @@ public struct StatusDetailView: View {
     .navigationBarTitleDisplayMode(.inline)
   }
 
+  @ViewBuilder
   private func makeStatusesListView(statuses: [Status]) -> some View {
-    ForEach(statuses) { status in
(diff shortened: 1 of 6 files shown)


Merrit:
Shame this hasn't been reviewed in 10 months, it would be a nice improvement.

Merrit:
Was this implemented elsewhere, or just closed?

Issue #2131, opened 2024-07-16
Where do downloaded images go on the desktop app?


datawench:
I've been using the ipad version for quite a while now, and am used to it. But I only recently decided to start using the desktop app. I see that when viewing an image I can use the "share" button to add it to the Photos library, but what does the little down arrow do, that I suppose must be a download button?

datawench:
I've edited the title, because I've tracked this down a little and I think it's an issue.
If I understand correctly, the "down arrow" on the popup image viewer uses the Swift function `UIImageWriteToSavedPhotosAlbum()` within the custom function `saveImage()` in the file [Packages/MediaUI/Sources/MediaUI/MediaUIView.swift](https://github.com/Dimillian/IceCubesApp/blob/a72f2900385dda2a52b70f40724060d33578b4ef/Packages/MediaUI/Sources/MediaUI/MediaUIView.swift#L185).

According to the documentation of the Swift function, it's supposed to write to the "camera roll" library, which on a Mac is just Photos. But it doesn't happen here. We can use the _share_ button to get the photo into the Photos library. We can even drag/drop the photo out of the popup to anywhere we want (a happy discovery). But as far as I can tell, the download feature just doesn't work on a Mac... though I know it does in iOS.

PR #2215, opened 2024-10-23
MediaUIView - Fix image download bug (#2131)

Taufi:
If the user has not yet granted permission to download photos, they will be asked to do so: `PHPhotoLibrary.requestAuthorization(for: .addOnly)`. Afterward, the photos can also be downloaded to the desktop. Previously, broader access was granted (`requestAuthorization(_:)`), but this method is now deprecated. Apparently, Mac Catalyst support for more advanced permission checks is still not fully developed. Once permission is denied, it can no longer be granted in the settings. You have to do it via a terminal command: `'tccutil reset PhotosAdd <your.app.bundle.identifier>'`.

### DIFF
--- a/IceCubesApp.xcodeproj/project.pbxproj
+++ b/IceCubesApp.xcodeproj/project.pbxproj
@@ -1545,7 +1545,7 @@
 				INFOPLIST_KEY_LSApplicationCategoryType = "public.app-category.social-networking";
 				INFOPLIST_KEY_NSCameraUsageDescription = "Upload photos & videos to attach to your Mastodon posts.";
 				INFOPLIST_KEY_NSHumanReadableCopyright = "© 2024 Thomas Ricouard";
-				INFOPLIST_KEY_NSPhotoLibraryAddUsageDescription = "";
+				INFOPLIST_KEY_NSPhotoLibraryAddUsageDescription = "Ice Cubes would like to save the selected photo in your photo library.";
 				INFOPLIST_KEY_NSPhotoLibraryUsageDescription = "Upload photos & videos to Mastodon";
 				INFOPLIST_KEY_UIApplicationSceneManifest_Generation = YES;
 				"INFOPLIST_KEY_UIApplicationSceneManifest_Generation[sdk=iphoneos*]" = YES;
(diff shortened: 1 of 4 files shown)


Packages/MediaUI/Sources/MediaUI/MediaUIView.swift
```diff
@@ -180,11 +183,19 @@ private struct SavePhotoToolbarItem: ToolbarContent, @unchecked Sendable {
     }
     return nil
   }
-
+  
   private func saveImage(url: URL) async -> Bool {
```
prodano-dev, line 187: Is it possible to change the var status into a let and use a switch case? Or is the if statement necessary after the requestAuthorization for addOnly? I had something in mind like this:

``` swift
private func saveImage(url: URL) async -> Bool {
        guard let image = try? await uiimageFor(url: url) else { return false }

        let status: PHAuthorizationStatus
        switch PHPhotoLibrary.authorizationStatus(for: .addOnly) {
        case .authorized:
            UIImageWriteToSavedPhotosAlbum(image, nil, nil, nil)
        default:
            await PHPhotoLibrary.requestAuthorization(for: .addOnly)
            status = PHPhotoLibrary.authorizationStatus(for: .addOnly)
        }
        return true
    }
```

Keeps it more readable without the if statements

IceCubesApp.xcodeproj/project.pbxproj
```diff
@@ -1545,6 +1545,7 @@
 				INFOPLIST_KEY_LSApplicationCategoryType = "public.app-category.social-networking";
 				INFOPLIST_KEY_NSCameraUsageDescription = "Upload photos & videos to attach to your Mastodon posts.";
 				INFOPLIST_KEY_NSHumanReadableCopyright = "© 2024 Thomas Ricouard";
+				INFOPLIST_KEY_NSPhotoLibraryAddUsageDescription = "";
```
ankushkushwaha, line 1548: Are we going to show empty description?

ankushkushwaha: Okay thanks for the explanation

Issue #2324, opened 2025-09-24
Bug: boost icon not clickable on an unlisted post when post is opened or viewed in a thread


Outer7418:
## Environment:

 - OS: iOS v26.0
 
 - IceCubesApp version: IceCubesApp v2.0.4
 

## Description
Trying to boost an unlisted post from a comment thread does not work correctly. Boosting from the Home view for example there is no issue but if the user opens the post or the thread and tries to boost, it does not trigger the "boost or quote" -selection. Instead, depending where the user is attempting the boost from, it either does nothing, or it opens the post.

Long pressing the post does open the context menu though and it is possible to boost that way though so there is a workaround at least.

## Related Issues

 - [x] Search that this bugs don't already exist before creating it.

Outer7418:
Here's a recording that hopefully shows this issue a bit more in detail. It might not be immediately obvious but at the beginning I'm trying to click the boost icon a few times but it isn't doing anything.

https://github.com/user-attachments/assets/3ddbe8d2-4cf5-4a5a-8556-63c29ae8f1e9

Dimillian:
Well the fix is actually to disable boost on unlisted post, like private posts, it's not supported on mastodon to boost unlisted posts

Outer7418:
> Well the fix is actually to disable boost on unlisted post, like private posts, it's not supported on mastodon to boost unlisted posts

I think that doesn't sound correct. At least according to the documentation I found it says that the unlisted (or quiet public as it's known currently it seems) is effectively the same as a public post but with only two differences:

- Your post will not appear in Mastodon’s live feeds or explore features.
- Your post will not appear in search results, even if your account has opted-in to full text search.

For private posts (edit: specifically meaning followers only -posts in this case, not mentioned users only -posts) it also actually says that user is able to boost their own private posts.

I'm not sure if I have the correct documentation here, but this is where I'm reading from:
https://docs.joinmastodon.org/user/posting/#unlisted

PR #2335, opened 2025-10-09
Disable boosting from context menu for non-public statuses

Dimillian:
## Summary
- disable the status context menu boost action when the underlying status cannot be boosted
- keep quoting availability unchanged while still allowing boosts of your own private posts

## Testing
- not run

------
https://chatgpt.com/codex/tasks/task_e_68e7775e1920832586c7a09d0c3f2cca


Dimillian:
Fix #2324

hazycora:
The boosting buttons do not look disabled visually (see attached screenshots), so it's completely unclear to the user why they can't boost the posts, but more importantly, unlisted posts *should* be able to be boosted, not just global posts. 
IMG_0233
IMG_0234

PR #2394, opened 2025-12-28
Update SC localization

Jerry23011:
Thank you Xcode :)


Dimillian:
what xcode is drunk again 
image

Jerry23011:
I noticed there are some conflicts, I'm converting this to draft first and will work on this later.

xabirequejo:
I have this same issue. I've tried to submit my changes three times already and they get rejected because of this. Apparently my file shows that I've deleted everything and written everything from scratch, when the reality is that I've just added missing strings or updated old one FOR A SINGLE LANGUAGE.

I'd like to find a solution, otherwise I'm just wasting my time.

shuang886:
The new version of Xcode appears to do two things:
1. Insert a space after the name of the data field. That is, what used to be:
```
"state": "translated",
```
got rewritten as:
```
"state" : "translated",
```
This can be fixed with a simple script after translation (**but please back up your edited Localizable.xcstrings first!**):
```
sed -i '' -E "s/\"(.*)\" \:/\"\1\"\:/" IceCubesApp/IceCubesApp/Resources/Localization/Localizable.xcstrings
```

2. The translations now appear to be sorted, which seems to mainly affect German (de).

My suggestion is for @Dimillian to generate a new copy of Localizable.xcstrings using Xcode 26(.2?) and require translators to upgrade. Problem 1 is not too tedious with the script, but problem 2 creates so many false positives in the diff that it obscures the actual changes.

Issue #2227, opened 2024-11-24
Feature request: Slightly altered UI / localization for filtered notifications


markasbach:
Description
-----------

On the Mastodon web ui, I've set "Unsolicited private mentions" to "Filtered" (see screenshot below). In Ice Cubes, this leads to a UI with two buttons that are not localized and not as easy to understand as in the web UI if such unsolicited private mentions actually happen. I would be very happy to see the same options with the same (translated) button texts in the app.

Notably, the English texts "accept" and "decline" on a page that shows filtered results, raises the question (in me, not a native speaker of English) if I actually want to "accept" *the filtering* or *the message/notification* - which are obviously very contrary things.

<img width="810" alt="screenshot_notification_settings_web" src="https://github.com/user-attachments/assets/86f1dca4-58cb-4630-af1e-b853c2e6c841">
<img width="696" alt="screenshot_notificication_actions_web" src="https://github.com/user-attachments/assets/d09f0001-a868-4a8a-9c01-9cff4a945373">
<img width="225" alt="screenshot_notification_actions_icecubesap" src="https://github.com/user-attachments/assets/bbf8c312-a14e-455d-8d11-5315e0b41aec">

Related Issues
--------------

- #2211 - potentially related, but only loosely
- #2210 - potentially related, but only loosely

Issue #2211, opened 2024-10-13
Bug: Filtered notifications are displayed even though setting is to hide completely


djfoheso:
## Environment:

 - OS: iOS v17.7
 
 - IceCubesApp version: IceCubesApp v1.11.1
 

## Description

I have set a filter via the Mastodon web client, selecting all Filter Contexts (including notifications), and setting the Filter Action to Hide Completely. In the web client this acts as expected: I don't see notifications matching the filter (a bunch of Japanese spam), and I don't get an alert that there are unread notifications.

In Ice Cubes, the filtered notifications appear in the Notifications panel, but they are greyed. In addition, I get a notification alert (the Notifications icon shows unread notifications).

Expected behavior: like the web client. The notifications should not be displayed at all, and I should not get a notification alert.

## Related Issues

 - [ ] Search that this bugs don't already exist before creating it.

Issue #2210, opened 2024-10-08
Bug: "Filtered Notifications" persists after disabling setting


ShadSterling:
## Environment:

 - OS: 
 
iOS 17.6.1
 - IceCubesApp version: 
 
1.11.1

## Description

The other day you added a "Filtered Notifications" entry pinned to the top of my notifications.  I don't want that filtering, and eventually managed to find the "Notifications Filter" settings where only "Moderated accounts" was set to something other than "Accept".  I changed the setting to "Accept", but the "Filtered Notifications" is still pinned to the top of my notifications

## Related Issues

 - [✔︎] Search that this bugs don't already exist before creating it.

Issue #639, opened 2023-02-04
Bug: Filter context not respected (post filtered where it shouldn't be)


csims:
Not a huge deal, but just something I noticed. I have a filter set up to hide matching posts completely, but only on home/lists and public timelines. I happened to make a post containing one of the words today, and noticed that Ice Cubes also hides the post completely on my profile. (I'm not sure how it would behave with the conversations and notifications options, since I haven't tested those out.)

Dimillian:
Ice Cubes do nothing fancy about filters, it just use the Mastodon API, so if the API don't send the post or send it filtered, it'll be filtered.

csims:
Interesting, ok! I had just compared to the web ui and another app I have installed (tooot), and the post was unfiltered on my profile with those. But maybe there's some other behavior at play there, I'm not sure!

Dimillian:
I'll take a look, maybe on your profile filter should be ignored? I thought the API would do that but might be wrong.

csims:
Just following up to add more context! Someone favorited & boosted my hidden post, and the notifications were filtered too (that option is also unchecked for my filter). It shows up in my notification list that someone favorited a post, but not what the post was

roydbt:
Still happens to me, on version 1.9.10. All my "Public timelines" filters still filter on my home timeline.

phpmacher:
@Dimillian 

I have to admit that I have this behavior too. That's why every time I see a notification without content, I have to go to another client and check what it contains.

IMHO this is to be considered a bug.

phpmacher:
I think this is still a pretty relevant bug. Also in the actual version.

Especially since you can't see why your own posts are suddenly missing from a thread.

In addition, I would suggest that if posts have been hidden in a thread, they are replaced by a reference to the affected filter.

phpmacher:
@Dimillian IMHO, this is one of the biggest bugs I encounter dozens of times every day when using the software.
In my opinion, this is still critical.

Apparently, there are other issues in this context that describe different behavior but point in the same direction: #2211.

It is certainly worth devoting some AI energy to this topic. ;-)

PR #2416, opened 2026-01-06
Fix filter context handling for notifications and timelines

Dimillian:
## Summary
- Make instance-side filter contexts authoritative across the app by mapping views to the correct context and matching filters by context (home/public/account/thread/notifications).
- Ensure hide-completely notifications never render while still advancing the notifications marker so unread badges stay accurate.
- Refresh policy + list after closing the notifications filter sheet so policy changes apply immediately.
- Localize the filtered-notifications selection UI and policy subtitles (new localization keys).

## Details by issue
- #639: Fixes context mismatch so filters only apply where configured (e.g., home/list vs profile/thread), preventing over-filtering on account/profile views and threads.
- #2211: Filters with “hide completely” no longer show in the notifications list and no longer trigger unread badge inflation; marker still advances.
- #2210: Closing the policy sheet now refreshes policy + list to remove the pinned filtered-notifications header when policies are set back to Accept.
- #2227: Replaces hardcoded English actions/subtitles with localized keys for the filtered-notifications UI.

## Testing
- mcp__XcodeBuildMCP__build_sim

### DIFF
--- a/Packages/Timeline/Sources/Timeline/View/TimelineViewModel.swift
+++ b/Packages/Timeline/Sources/Timeline/View/TimelineViewModel.swift
@@ -135,6 +135,9 @@ import SwiftUI
 
   init(statusFetcher: TimelineStatusFetching = TimelineStatusFetcher()) {
     self.statusFetcher = statusFetcher
+    Task {
+      await datasource.setFilterContext(timeline.filterContext)
+    }
   }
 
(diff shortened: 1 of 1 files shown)


Dimillian:
@codex review this

Dimillian:
@codex review this in accordance with the mastodon instance side filter documentation and see if it honors it correctly

Issue #2383, opened 2025-12-15
Bug: screen rotation gets stuck rendering app unusable


jonathankoren:
## Environment:
 - OS: iOS 26.1
 - IceCubesApp version: 2.0.9
- iPhone 17

## Description
If the screen rotates from portrait to landscape and back while the compose screen is active, it can get stuck with less than half the screen actually being visible.

This is a long standing bug on iPhone. Furthermore, this bug seems doubly wrong because according to issue 670, this functionality shouldn’t even exist on iPhone. 

Screenshot attached 

Replicability 80%

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/378ae2e7-000d-44e3-b5e9-dfb2bf04804a" />

## Related Issues
https://github.com/Dimillian/IceCubesApp/issues/670

 - [ ] Search that this bugs don't already exist before creating it.

ppuschmann:
Can reproduce with version 2.1.3.

Additional infomation:
- no further editing is possible
- going to homescreen and back to the app only hides the keyboard
- no further action is possible (not even "save draft" or send), and yes, even before going back to the homescreen
- rotatating into the "faulty" landscape-mode does not resume editing
- the app will crash shortly after

iPhone 11 Pro Max, with iOS 26.2.1

<img width="621" height="1344" alt="Image" src="https://github.com/user-attachments/assets/5eca0510-d225-4b24-9daa-6e7d033efb69" />

<img width="1344" height="621" alt="Image" src="https://github.com/user-attachments/assets/f650b5f9-f8eb-4ccf-a7fc-8c8f8a7562eb" />

Issue #670, opened 2023-02-06
Feature request: allow for screen orientation depending on rotation of device.


ScottChampagne:
If I switch from holding my iOS device vertically to a horizontal orientation, the screen should reflect that behavior. (And vice versa)

PR #2418, opened 2026-01-06
Disable landscape orientation on iPhone

ghall89:
Addresses #2383 

Disable landscape orientations on iPhone, since it seems [landscape support is not planned](https://github.com/Dimillian/IceCubesApp/issues/670#issuecomment-1418834967) for iPhone.

### DIFF
--- a/IceCubesApp.xcodeproj/project.pbxproj
+++ b/IceCubesApp.xcodeproj/project.pbxproj
@@ -1064,6 +1064,7 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = IceCubesApp/Info.plist;
 				INFOPLIST_KEY_CFBundleDisplayName = "Ice Cubes";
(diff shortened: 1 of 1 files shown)


stepheneb:
Will this disable landscape orientation on iphone when viewing an image?

ghall89:
@stepheneb That's a good point, it would disable landscape when viewing an image. I hadn't considered that, but that would probably be a better experience than disabling rotation on iPhone altogether.

stepheneb:
There are many times when viewing a screenshot someone has shared I need to both switch to landscape and zoom in so see details, Am running on a iphone with a smaller screen.

stepheneb:
One other use I have made of landscape mode. When someone includes a url in their post that provides andimage -- and the image has detail that is hard to make out and I'd like to read more of the image before deciding to open the url I rotate the phone to see the image better. I find my self often wanting to just tap on the image to view an expanded view without rotating but a regular tap of course opens the url. Maybe something like "Open Image" could be added to the long tap contextual menu.

PR #2419, opened 2026-01-06
Add functionality to hide posts created by bots

jappiedev:
This adds the functionality to hide posts created by bots. 

I tried to add the localization string, but it created a lot of changes.


jappiedev:
I saw that you updated the localization file. I've added translations now.

There wasn't an issue for this, but I added this feature to scratch my own itch.

There is also an issue and PR for Mastodon itself, so I think there are more people with this itch:

https://github.com/mastodon/mastodon/issues/8293
https://github.com/mastodon/mastodon/pull/11272

PR #2434, opened 2026-01-15
increase visual contrast in poll ui

stepheneb:
... so the text is more easily readable.

- poll bars w/results
- Send Vote button

IMG_4235


stepheneb:
Realized the "Send Vote" was hard to read from lack of visual contrast also.

stepheneb:
When checking other standard themes this is going to take more work and thinking.

Examples with issues:

- Desert - Light
  - image
  - image

- Ice Cube - Light
  - image
  - image

PR #2437, opened 2026-01-25
Switch from dispatch queue to main actor in a task

LumioseSil:
This is so we can prevent race conditions after an arbitrary time too.


AZero13:
Also claude and co suggesting to avoid timing issues by having async after an arbitrary period of time just... why?!!!!

PR #2473, opened 2026-07-08
Fix crash on rotation while drafting a post (#2461)

### LINKED ISSUE #2461: Bug: screen rotation while drafting crashes app
## Environment:

 - OS: 
 iOS 26.4.2
 - IceCubesApp version: 2.1.3
 

## Description

If screen orientation rotates while drafting a post the app will crash.

meet4731:
Fixes #2461 — the app crashes if you rotate the device while the post composer is open.

Digging into it (and matching the AttributeGraph cycle log pylapp posted on the issue), the crash is a SwiftUI AttributeGraph cycle: when a focused UITextView is in the compose sheet and the interface rotates, the update loop cycles and the watchdog kills the app. The existing device-orientation observer doesn't help here because it only fires after the rotation has already happened, by which point the cycle's formed.

The fix is to host the editor's text view in a small TextViewController (UIViewControllerRepresentable) so it gets a viewWillTransition hook. That fires just before the rotation animation starts, so resigning first responder there drops the keyboard in time to stop the cycle forming in the first place. I've left the orientation observer in as a backstop for the post-rotation case.

One extra change: I had to qualify SwiftSoup.Document in HTMLString.swift — it was hitting an ambiguous type-lookup error building on the current toolchain.

Tested by reproducing the crash (open composer, rotate) and confirming it's gone. Editor sizing and behaviour are unchanged; the only difference is focus now starts on tap rather than automatically.


meet4731:
Hi. Dimillian. Waiting for you to review it. Thanks

PR #2477, opened 2026-07-30
New Bluesky theme set

jadetheda:
Pulled the exact colors from the website’s source code, and I think it looks pretty cool!
IMG_1016
IMG_1017
IMG_1018


jadetheda:
I was split on which color to choose for the dark theme’s tint color.

The tint color affects both link color and the color of UI elements like the active tabs (Home, Timeline, etc), but in the dark mode for Bluesky these are different colors, with UI elements being a brighter blue.

I ended up going with the color of links (#4D97FF) for it, but if we wanted to make it the color of the UI elements, it would be #0F73FF (used by both the Compose button and the Follow button).

IMG_1024
IMG_1027
IMG_1022

